PR #5, opened 2020-04-18
Add failing test for shorthand conditional compose(cond && token)

CodingDive:
I couldn't find the correct place to fix the test. It would be cool if this was supported without having to `ts-ignore`

### DIFF
--- a/packages/css/tests/index.test.ts
+++ b/packages/css/tests/index.test.ts
@@ -169,6 +169,18 @@ describe("createCss", () => {
       String(css.compose(undefined, null, false, "", css.color("red")))
     ).toBe("c_0");
   });
+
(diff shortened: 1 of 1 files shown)


packages/css/tests/index.test.ts
```diff
@@ -169,6 +169,18 @@ describe("createCss", () => {
...
   });
+
+  test("should allow empty compose call", () => {
+    const css = createCss({}, null);
+    expect(css.compose()).toBe("");
+  });
+
+  test("should allow conditional compositions", () => {
+    const css = createCss({}, null);
+    expect(String(css.compose(false && css.color("red")))).toBe("");
```
CodingDive, line 180: typescript screams at me here 😭

PR #50, opened 2020-07-08
@stitches/tailwind: update @stitches/css version

kingdaro:
Before, tailwind was including its own (incompatible) version of css, which was causing a lot of typing issues in a local environment. I had to add `"resolutions": { "@stitches/css": "3.0.2" },` to my package.json to fix this

Also moves it to use a peer dep, which seems more correct in this case, since it's likely a mistake if `@stitches/css` isn't installed explicitly

### DIFF
--- a/packages/tailwind/package.json
+++ b/packages/tailwind/package.json
@@ -39,10 +39,10 @@
     "theme"
   ],
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


peduarte:
Hi, the tailwind package is now deprecated. @christianalfoni might continue this in his own repo thought 👍

PR #56, opened 2020-07-15
Fix PolymorphicComponentProps inference

jjenzz:
I'm not entirely sure if the code I've removed is needed for an edge-case but without this change, prop types were breaking. 

**BEFORE**

types

**AFTER**

types


peduarte:
@jjenzz @christianalfoni do we still need this?

PR #73, opened 2020-08-05
Add simple ampersand support and fix nesting normal alphabet selectors

### LINKED ISSUE #60: Add ampersand support for selectors
Our primitives would export something like:

```jsx
export const resetStyles = {
  root: {
    display: 'flex',

    '&[data-state-disabled]': {
      pointerEvents: 'none',
    },
  },
};
```

and then the consumer would do:

```jsx
import { ComponentPrimitive, resetStyles } from '@interop/primitives';

const Component = styled(ComponentPrimitive, {
  ...resetStyles.root,
  border: '1px solid green';
});
```

The ampersand means this spreading by the consumer would work in styled-components, emotion and also stitches.

@christianalfoni was thinking we could have selector logic as:

- If it starts with `&`, remove it
- If it now starts with `:` or `[`, treat as state selector
- If not, insert a space to treat as nested element selector

```jsx
{
  '&:hover': {},
  'div': {},
  '&[..]': {},
}
```

hadihallak:
### A couple of small changes:
---------
**Adds simple ampersand support (Fixes #60):**
Before: 
```js
const Button = styled.button({
  ':hover': {},
  ' .nested': {},
})
```

After:
```js
const Button = styled.button({
  '&:hover': {},
  '.nested': {},
})
```

> Note: as agreed on slack and in #60. this is a simple implementation that only deals with the ampersand at the start of a selector and doesn't try to mock the way it works in Sass and other css in js libraries.

---------
**Fixes nesting for rules that start with an alphabetical character (span, div, etc..)**
Before:
```js
const Button = styled.button({
// span styles would be treated as a single css value
  'span': {},

})
```

After:
It will treat it as a nested style object correctly.

### DIFF
--- a/packages/css/src/index.ts
+++ b/packages/css/src/index.ts
@@ -347,7 +347,7 @@ export const createCss = <T extends IConfig>(
           );
         }
         createCssAtoms(props[prop], cb, prop, pseudo);
-      } else if (!prop[0].match(/[a-zA-Z]/)) {
+      } else if (typeof props[prop] === 'object') {
         createCssAtoms(props[prop], cb, screen, pseudo.concat(prop));
       } else if (canCallUtils && prop in utils) {
         createCssAtoms(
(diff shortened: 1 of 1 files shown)


hadihallak:
Closing since this will be resolved in a different manner in the keyframes PR.

PR #74, opened 2020-08-05
Prevent css prop from appearing on the dom

hadihallak:
Before this fix, the css prop would get passed to the actual dom element so this fixes it so that it doesn't show up on the dom just like variants.


peduarte:
Looks like @christianalfoni has done this as part of his `pocs`, so maybe we can close this PR

@christianalfoni can you confirm?

Issue #76, opened 2020-08-06
Consider compressing hashed atomic classnames


peduarte:
Currently they are: `brw_3001088182 bbw_115517177 blw_1917871437 bts_1515315272`, wonder if we can keep them shorter in production?

christianalfoni:
@hadihallak asked about the hashing algorithm, if there was any special reason for using the one we use. Nopes, we can use something else which produces shorter names 😄 

I believe styled-components has a pretty nice one, we could just use that. 

https://github.com/styled-components/styled-components/blob/master/packages/styled-components/src/utils/hash.js

Issue #75, opened 2020-08-06
Improve testing


peduarte:
We need to put in place a better release flow to avoid regressions.

Im not quite sure whats the best way to do this, but open to suggestions @christianalfoni and @hadihallak 

Ideally, Id like to have a test suite of all the things we know we need to support. Im happy to add the tests once the structure is in place

christianalfoni:
What is missing is running React tests with JSDOM ideally, so that we can actually test the injection. So Jest is already set up, but need to add something like Enzyme I believe. Here is a tutorial: https://medium.com/@salihunuruddeen/how-to-configure-jestjs-unit-test-environment-for-test-in-react-js-7b8b6b9a2a0b

For the typing it is required to set up a different kind of test suite: https://www.simonholywell.com/post/testing-typescript-types.html. Never set up that before, but yeah, here is an example 😄

PR #88, opened 2020-08-12
Feature/css keyframe animations

### LINKED ISSUE #60: Add ampersand support for selectors

### LINKED ISSUE #68: Add keyframe animation
We need to support keyframe animation 🙌

hadihallak:
**Main key points:**
- Keyframes support resolves #68  :
Abstracted the utils/specificity props/screens/ selectors resolving mechanism into functions so that the internal implementation could be shared with `keyframes` and soon in `createGlobalStyles`.
`createCssAtoms` & `createUtilsAtoms` got replaced with the above implementation the same as `keyframes`
- Altered hashing function to produce smaller size classes (needed to hash animations into an animation-name) (fixes: #76) 
- Adds simple ampersand support (Fixes #60)
- Updated tests to work with the new ampersand syntax. @peduarte turns out that christianalfoni already had some tests in place for the CSS package which helped me a lot with discovering bugs I introduced with the abstraction above so might be a good place to start if you want to add to the existing tests in there until we setup testing properly in #75. see `css/tests`
- Longhand parsing will be added in a separate PR after I do a separate package for the whole parsing shenanigans sometime next week

### DIFF
--- a/packages/css/src/index.ts
+++ b/packages/css/src/index.ts
@@ -8,6 +8,7 @@ import {
   ISheet,
   IThemeAtom,
   ITokensDefinition,
+  IKeyframesAtom,
   TCss,
 } from "./types";
 import {
@@ -22,6 +23,11 @@ export * from "./types";
 export * from "./css-types";
 
 export const hotReloadingCache = new Map<string, any>();
+const hyphenCssProp = (cssProp: string) =>
+  cssProp
+    .split(/(?=[A-Z])/)
+    .map((g) => g.toLowerCase())
+    .join("-");
 
 const toStringCachedAtom = function (this: IAtom) {
   return this._className!;
@@ -144,6 +150,17 @@ const createThemeToString = (classPrefix: string, variablesSheet: ISheet) =>
     return themeClassName;
   };
 
+const createKeyframesToString = (variablesSheet: ISheet) =>
+  function toString(this: IKeyframesAtom) {
+    if (this._cssRuleString) {
+      variablesSheet.insertRule(this._cssRuleString);
+    }
+
+    this.toString = () => this.id;
+
+    return this.id;
+  };
+
 const composeIntoMap = (
   map: Map<string, IAtom>,
   atoms: (IAtom | IComposedAtom)[]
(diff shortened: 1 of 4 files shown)


peduarte:
Did some quick tests:

- `css.keyframes()` seems to be working well ✅ 
- `&:hover` is working well too
- `:hover` (without `&`) has stopped working, this should still work. Here's an [example in SC](https://codesandbox.io/s/stoic-sun-n2ykf?file=/src/App.js)

@hadihallak on a separate note, I realised the `animation` shorthand property doesn't get compiled to its longhand counterparts. Thought id just let you know. Once the parser work is done, would that happen?

CleanShot 2020-08-12 at 11 49 48@2x

hadihallak:
> :hover (without &) has stopped working, this should still work.

should be fixed now

-------

> @hadihallak on a separate note, I realised the animation shorthand property doesn't get compiled to its longhand counterparts. Thought id just let you know. Once the parser work is done, would that happen?

once we implement the parser this will work as expected:
**here is a sandbox with the custom parser for the animation prop if you want to take a peek:**
https://codesandbox.io/s/parse-values-xq61v?file=/src/App.tsx
currently, it matches most (if not all) valid animation shorthand syntax
the internal tokenizer also needs some work to handle quoted values and slashes but it's already good enough for animation

-------

I've also added tests for the keyframes method and commented the code a bit more.

peduarte:
@hadihallak nice one. I did some tests and everything seems fine.
@christianalfoni would appreciate if you could take a look over, since this is Abduls first PR. Cheers

PR #91, opened 2020-08-13
Fix borderL pointing to borderB in Tailwind

geirsagberg:


peduarte:
Hi, the tailwind package is now deprecated. @christianalfoni might continue this in his own repo thought 👍

Issue #103, opened 2020-08-18
Extra Shorthands


colmtuite:
I think we forgot to mention shorthands for `flex` and `flex-flow`.

peduarte:
Also:
- `grid`
- `overflow`
- `outline`
- `list-style`

richardbruskowski:
Not really sure if I simply did something wrong on my side, but couldn't get these to work, while `border-top` etc. worked fine, maybe they are missing as well:
- `border-block-start`
- `border-block-end`
- ...

rafgraph:
Tokens in outline shorthand are not working for me, e.g. `outline: '2px solid $green600'`

peduarte:
We've published a pre-release of Stitches beta 🚀

**Core**:
```bash
npm install @stitches/core@0.1.0-canary.0
```

**React**:
```bash
npm install @stitches/react@0.1.0-canary.0
```

This issue should have been resolved in the beta version. We'd appreciate it if you could verify it. 

⚠️ Please refer to this [migration guide](https://stitches-site-git-beta.modulz-deploys.com/blog/migrating-from-alpha-to-beta) as there are some breaking changes.

PR #105, opened 2020-08-18
Feature/initial shorthand parsers

### LINKED ISSUE #58: `border: none` 💥
I've noticed that the following declaration:
```
border: 0
```

yield some unexpected results. (@christianalfoni  I think you came across that in yesterday's stream too)

I kinda understand _why_ this may happen, but think it's an unintuitive DX. 

Changing from `0` to `'0'` (String) partially solves the issue, but some of the border long-hands property are still being generated as `undefined`:

CleanShot 2020-07-29 at 10 58 52@2x

hadihallak:
**What's inside:**
- Migrate the initial parsers to the main repo
- added the string-based parser and striped it from some functionality to make it faster
- Fixed a couple of bugs in there
- Wrote unit tests for the parsers and some integration tests to make sure that they're wired correctly to the css package
- Updated the current tests for the changes

Note: the parsing was actually very good but I found a few bugs which I left as commented tests to get to in the next iteration for #103 
Also, we seem to not be handling default values which is something that I noticed that the browser prefills longhand versions with when shorthand was supplied without enough properties

Fixes #58


packages/css/tests/value-tokenizer.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    ]);
+  });
+
+  test("Handles nested css functions in multiple groups", () => {
+    expect(
+      tokenizeValue("linear-gradient(red, hsla(120,100%,50%,0.3)), red")
+    ).toEqual([["linear-gradient(red, hsla(120,100%,50%,0.3))"], ["red"]]);
+  });
+
+});
```
peduarte, line 55: Is it worth adding some tests to check that `/` work? like for `font` or `borderRadius`?

peduarte (approved):

packages/css/tests/value-tokenizer.test.ts
hadihallak: I believe it works and there are tests for it at the upper level in the shorthand parsers tests. 
but yea I agree, if we break the parsing functionality in the future, catching the error here would be preferable 👍 

will add a case for this.

peduarte (approved):

PR #108, opened 2020-08-19
require tokens to be used when defined

fgnass:
With this change the compiler will yield an error when tokens are defined but raw values are used instead.

carbon

In order to use non-token values developers must add `as any` or `@ts-ignore`.
 
See https://twitter.com/fgnass/status/1296034855562743808 for the discussion with @christianalfoni

### DIFF
--- a/packages/css/src/types.ts
+++ b/packages/css/src/types.ts
@@ -49,7 +49,7 @@ export type TRecursiveCss<
   T extends TConfig,
   D = {
     [K in keyof Properties]?: K extends keyof ICssPropToToken<T>
-      ? ICssPropToToken<T>[K] | Properties[K]
+      ? ICssPropToToken<T>[K]
       : Properties[K];
   }
 > = (
@@ -68,7 +68,7 @@ export type TFlatCSS<
   T extends TConfig,
   D = {
     [K in keyof Properties]?: K extends keyof ICssPropToToken<T>
-      ? ICssPropToToken<T>[K] | Properties[K]
+      ? ICssPropToToken<T>[K]
       : Properties[K];
   }
 > = D;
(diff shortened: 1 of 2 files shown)


peduarte:
Hey @fgnass, thanks for the PR ✌️ 

I think enforcing the usage of tokens is a great idea to push towards a more consistent and systematic product. 

Our goal with Stitches is to provide the best developer experience. As part of that, we'd like the library to be as least intrusive as possible. 

I think at this early stage, I'd prefer to keep it as open as possible. But let's hear from @colmtuite @christianalfoni @hadihallak 

I'd be fine if this was an opt-in type of thing. Something like `strict: true` in the config.

colmtuite:
Thanks for the suggestion @fgnass.

Past a certain team or project size, I'm on board with enforcing tokens. But I don't think it should be the responsibility of this library. It should be a manual extra step, or handled at code review.

The reason is that I don't think tokens are the only viable approach to styling. There are plenty of very small teams/projects who don't necessarily benefit much from tokens at all, and it's perfectly fine.

So even though it would be optional and there would also be an escape hatch, I don't think it belongs as a feature of this lib.

fgnass:
My reasoning was that teams who don't want or need tokens probably also wouldn't define them in the first place. If no tokens are defined the compiler will happily accept any value without complaining. Once you define tokens you opt in to using them and you are rewarded with an unmatched DX as your IDE now auto-suggests the list of valid values.

Alternatively it would be pretty easy to add a `strict: true` setting like @peduarte suggested it. Maybe that could be the way to get best of both worlds?

peduarte:
@fgnass there's an ongoing discussion related to this here https://github.com/modulz/stitches/issues/114#issuecomment-677682663

I've heard from a lot of people today that want to be shouted that when they're using an invalid token, so clearly this is a requested feature. But yeah, we need to make sure that by default its a bit looser. 

Would you like to give it a go at implementing the `strict` functionality?

peduarte:
Can you please bring these changes into https://github.com/modulz/stitches/pull/116

Thanks!

PR #109, opened 2020-08-19
Allow incomplete tokensets

fgnass:
This is a followup to #108 that allows the use of _non-token_ values if no tokens are defined for the corresponding token-category. See the included unit test for details.


peduarte:
I left a comment in #108 

Let's see what happens there first 👍

peduarte:
Is this PR still needed, or does #116 cover this too?

fgnass:
Closing this in favor of #116 which also includes this fix.

Issue #86, opened 2020-08-11
Media queries inside variants


colmtuite:
We need to be able to place media queries inside a variant block.

```
.Button.blue {
  @media (hover: hover) {
    // cool hover effect for the blue variant
  }
}
```

This is possible in all popular css-in-js libs and preprocessors.

Without it, styling components becomes unmaintainable.

This may be related to #79.

peduarte:
Yep, definitely 💯 

Here's a Codesandbox with a repro: https://codesandbox.io/s/stitches-media-query-in-variant-sd93t?file=/src/App.tsx

As a rule of thumb: anything that can be done in the base styles, should be supported in variants

/cc @christianalfoni just wondering if you had ever come across this and if it was intentional? (the error message is funny!)

hadihallak:
@peduarte been looking into supporting inline media queries today and @christianalfoni keep me honest but I think this might be intentional because currently, we don't support nesting screens and because variants get converted to responsive ones that map to your screens when you do a screen inside a variant you're practically nesting the screen in all other defined screens which is disallowed intentionally for reasons @christianalfoni  mentioned here https://github.com/modulz/stitches/issues/67#issuecomment-671025206

Still looking into this but overall, I'm in agreeance with @christianalfoni on the whole nesting media queries thing, as it's not only very tricky but, might be near-impossible to solve gracefully.

hadihallak:
so looks like Otion did the sorting thing to get around the specificity issues which appear to be working well from my initial testing so will give this another spin tomorrow and see how it goes

peduarte:
@hadihallak yeah I fully understand the complexity here 😕 

Personally, I'd be totally fine with not supporting media queries within the Variant Style object, and instead, force people to change their variants responsively at the consumption layer. But Im not sure everyone is ready to think this way yet...

Hmm, is there a way to know that, if a variant contains responsive styles, and the user users the responsive syntax to apply that variant, we throw a warning? (or ignore one of the media queries)? Just throwing an idea out there :D

colmtuite:
The issue with this one is not about breakpoints. With breakpoints, it's fine to just define your variants and then apply them at consumption.

This issue is about all of the other media queries which are not breakpoints. Like `@hover: hover`, for example, which should be applied every single time you define a hover pseudo class. Or `(prefers-reduced-motion)`, or any of the others.

[...]

Issue #83, opened 2020-08-10
:hover and :active specificity issue


peduarte:
When trying to add `:hover` and `:active` pseudo states in CSS, we'd usually do:
```css
button:hover {}
button:active {}
```

This is so when you mousedown while hovered, the `:active` style takes precedence. 

Adding the same rules, in the same order, in Stitches (running in CRA), yields unexpected result.

Repro: https://codesandbox.io/s/stitches-hoveractive-states-sezl2?file=/src/App.tsx:177-259

In the Codesandbox above, notice how when you mousedown the background color remains red. It should change to blue. 

Out of curiosity, I tried to do the same with SSR (Next.js), and the SSR generated styles works as expected 😄 

SSR Demo: https://codesandbox.io/s/stitches-hoveractive-states-with-ssr-lyc6f?file=/pages/index.js

hadihallak:
This is related to the insertion order of the css rules and I was able to re-produce in a nextjs app here: https://codesandbox.io/s/stitches-hoveractive-states-with-ssr-forked-4iq12?file=/pages/index.js

I think the inconsistency between the result of CRA and Next is unrelated but is worth digging into too so will open another issue for it once i know for sure what's going on

but back to this specificity issue. not sure how we would solve this but my initial thinking is that if it's ok then we hardcode some logic in to make it so that :active rules are always injected after :hover rules.

if the above is very limiting and we need the user to be completely in control of the precedence then the only thing i could think of is to create compound rules so if a user has 
```js
:hover {background: red},
:active {background: black} 
```

we then would resolve this into 3 rules
```js
:hover { background: red}
:active { background: black}
// active comes after hover so we create a compound atom with the background black
:hover:active { background: black}
```

however if the rules were in **reverse** then we'd do this:
```js
// we swapped the order of active and hover
:active {background: black} 
:hover {background: red},
```

we then would resolve this into 3 rules
```js
:active { background: black}
:hover { background: red}
// hover comes after hover so we create a compound atom with the background red
:hover:active { background: red}
```

I think this might work but things could get very hairy with many pseudo-classes

jjenzz:
We wouldn't be able to compound them because that would prevent active styles for keyboard users. I'm curious how other atomic libraries handle this. Have we taken a look at Fela's or Styletron's implementation here?

peduarte:
> my initial thinking is that if it's ok then we hardcode some logic in to make it so that :active rules are always injected after :hover rules

[...]

PR #112, opened 2020-08-19
Breakpoints, inline media queries, ampersand and pseudo presedence

christianalfoni:
Okay,

So this PR should contain:

1. Breakpoints as discussed here: https://github.com/modulz/stitches/issues/86#issuecomment-675590197
2. Ampersand support: https://github.com/modulz/stitches/issues/60
3. Pseudo presedence as mentioned here: https://github.com/modulz/stitches/issues/83#issuecomment-673942355

I wrote simple tests for these, but looks good so far 😄

### DIFF
--- a/packages/css/src/index.ts
+++ b/packages/css/src/index.ts
@@ -484,7 +484,7 @@ export const createCss = <T extends TConfig>(
     const inlineMediaQueries = selectors?.filter((part) =>
       part.startsWith("@")
     );
-    const selectorString = selectors
+    let selectorString = selectors
       ?.filter((part) => !part.startsWith("@"))
       .join("");
 
@@ -510,6 +510,16 @@ export const createCss = <T extends TConfig>(
       vendorProps,
       vendorPrefix
     );
+
+    // We want certain pseudo selectors to take presedence over other pseudo
+    // selectors, so we increase specificity
+    if (!selectorString?.match("&")) {
+      if (selectorString?.match(/\:focus|\:focus-within|\:hover/)) {
+        selectorString = `&&${selectorString}`;
+      } else if (selectorString?.match(/\:active/)) {
+        selectorString = `&&&${selectorString}`;
+      }
+    }
 
     // Create a new atom
     const atom: IAtom = {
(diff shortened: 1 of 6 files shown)


peduarte (approved):
LGTM, but I'd like to run proper tests when its in master. 

@hadihallak can you review?

hadihallak (commented):
Looks great 💯 
Thanks for implementing all of that against the updated master. very helpful.

This doesn’t implement the responsive atom warning logic we talked about in #86 right?

I think there might’ve been some confusion regarding breakpoints and media queries so apologies about that but the team agreed that mediaQueries is better suited than breakpoints even if we're gonna need to do some work internally separating breakpoints from nestable media queries.

@jjenzz also suggested a mixins idea which was very interesting and would solve the confusion we would have between breakpoints and media queries that we're likely to explore with later but until then the team agreed that mediaQueries provide a better experience so we're sticking with this.

In both cases, I think I could take it from here and implement this on top of this branch or if you’re still working here, I could do this in a separate branch as well.

Thanks again for putting this together 🙏

jjenzz:
In the currently released version of stitches the following was possible but breaks with this PR https://codesandbox.io/s/kind-kalam-1itou

hadihallak:
LIkely the product of #88 as it altered recession. will fix this and add a test for this use-case for the future

packages/css/src/index.ts
```diff
@@ -125,14 +125,17 @@ const processStyleObject = (
     // shorthand css props or css props that has baked in handling:
     // see specificityProps in ./utils
     if (isSpecificityProp) {
-      const resolvedSpecificityProps = specificityProps[key](config.tokens, val);
+      const resolvedSpecificityProps = specificityProps[key](
+        config.tokens,
+        val
+      );
```
peduarte, line 131: prettier

packages/css/src/index.ts
```diff
@@ -125,14 +125,17 @@ const processStyleObject = (
...
       processStyleObject(
         resolvedSpecificityProps,
         config,
         valueMiddleware,
         [...currentNestingPath],
         false,
-        false,
+        false
```
peduarte, line 138: prettier

hadihallak (approved):

christianalfoni:
Hi there!

Sorry for late reply, a bit busy today 😄 

But this looks great! Happy it is in!!

PR #116, opened 2020-08-20
Strict mode

fgnass:
As discussed in #108 this PR adds a strict flag to the config.

### DIFF
--- a/packages/css/tests/index.test.ts
+++ b/packages/css/tests/index.test.ts
@@ -608,7 +608,7 @@ describe("createCss", () => {
     );
   });
 
-    test("should inject styles for animations into sheet", () => {
+  test("should inject styles for animations into sheet", () => {
     const css = createCss({}, null);
     const keyFrame = css.keyframes({
       "0%": { background: "red" },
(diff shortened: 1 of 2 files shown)


packages/css/tests/index.test.ts
```diff
@@ -995,4 +995,48 @@ describe("createCss", () => {
...
+      strict: true,
+      tokens: {
+        colors: {
+          primary: "red",
+        },
+      },
+    });
+    css({
+      color: "primary",
+      // @ts-ignore
```
peduarte, line 1023: I saw in a diff PR you used `@ts-expect-error`. just wondering if its also worth using it here instead of a `@ts-ignore`?

peduarte:
@fgnass hey. sorry for the delay on this. we had to put it on hold, as we were getting ready to officially launch stitches.

I'm keen on getting this in for the next release. we need to:
- resolve conflicts
- bring the changes from https://github.com/modulz/stitches/pull/108 into this
- make sure there are tests
- set this PR against the `canary` branch, instead of `master`

Thanks!

Issue #119, opened 2020-08-20
SSR and Client side rendering inconsistencies regarding injection order. 


hadihallak:
in client-side apps, injection is always at index 0 where on SSR it appears to be pushing the rule at the end of the rules array

peduarte:
@hadihallak is this fixed?

hadihallak:
yep

Issue #64, opened 2020-08-03
Browser specific pseudo-elements error when used in other browsers


hadihallak:
Fixing it is probably as simple as wrapping the insertRule call in a try-catch block to avoid the browser complaining about an unsupported selector.

**Error message:**
```js
SyntaxError: Failed to execute 'insertRule' on 'CSSStyleSheet': Failed to parse the rule '.bc_0::-moz-range-track{background-color:transparent;}'.
toString
src/index.ts:65
  62 |     cssRule = `.${className[0]}{${this.cssHyphenProp}:${value};}`;
  63 |   }
  64 | 
> 65 |   sheets[this.screen].insertRule(
     | ^  66 |     this.screen ? screens[this.screen](cssRule) : cssRule
  67 |   );
  68 | }
```

**the following code will give the above error when used in chrome:**

```tsx

import React from 'react'

import { createStyled } from '@stitches/styled'

export const { styled, css } = createStyled({})

export const Slider = styled.div({
  '::-moz-range-track': { backgroundColor: 'transparent' },
})

function App() {
  return (
    <div className="App">
      <Slider> Working </Slider>
    </div>
  )
}

export default App
```

christianalfoni:
I believe you could also use this:

https://github.com/modulz/stitches/blob/master/packages/css/src/utils.ts#L281

Here you get a list of all the supported properties of the browser, so could check if it exists there. But yeah, then you would have to do that for every single property which would give some overhead. In support of the try catch here, but just wanted to mention it 😄

hadihallak:
fixed in #120

PR #120, opened 2020-08-20
Bug fixes

hadihallak:
**A few bug fixes:**
- Media queries are now guaranteed to be injected after normal css rules which is the behavior of other cssInJs libraries 
- Fix inconsistencies of rendering styles between the server and the client #119 
- Warns users if two different atoms that use the same inline media query, selector, breakpoint, and value combo that this will cause specificity issues https://github.com/modulz/stitches/issues/86#issuecomment-675491254
- fix inserRule bug #64 

Will do a bit more testing and cleanup then will merge but this should be ready for review


peduarte (approved):

peduarte:
Awesome stuff. Will prepare an alpha release for testing 🔥

PR #131, opened 2020-08-25
fix typing

peduarte:


peduarte:
Tested this locally and seems to be working! will release to test fully e2e

Issue #133, opened 2020-08-26
value autocomplete isn't working inside variants.


hadihallak:
You will get the autocomplete for the property but not the value when typing CSS inside a variant

spent a fair bit of time on this and it seems connected to the use of generics and the only way I could get it to work was if I break variants typing.

**Steps to reproduce:**
Create a styled component with variants then try typing something like `backgroundColor` inside a variant. notice how only the key is autocompleted and the value itself is not typed or autocompleted

hadihallak:
Fixed in #134

Armadillidiid:
I'm experiencing the same issue on version `0.6.1`

Issue #110, opened 2020-08-19
Error when trying to add a ref to a styled component


rubenmoya:
Hello!

I'm migrating a small project from emotion/styled-system to stitches and I'm having some problems with the `ref` attribute.

I've created [a codesandbox](https://codesandbox.io/s/stitchesstyled-ref-rv0m9?file=/src/App.tsx) if you want to have a look at the error, this is the code:

```tsx
import * as React from "react";
import { createStyled } from "@stitches/styled";

const { styled } = createStyled({});

const Box = styled("div", {});

export default function App() {
  const boxRef = React.useRef<HTMLDivElement>(null);
  const divRef = React.useRef<HTMLDivElement>(null);

  return (
    <div>
      <Box ref={boxRef}>Hello there</Box>
      <div ref={divRef}>General Kenobi</div>
    </div>
  );
}
```

The error seems to be `Property 'ref' does not exist on type [...]` (although the error text is huge), am I missing something? 🤔

Besides that problem I'm loving it so far :) much more cleaner than using emotion/styled-system 👏 I can't wait for the official release

peduarte:
Hey @rubenmoya thanks for raising this issue ✌️ We'll take a look.

Let me know how you get on with the migration!

rubenmoya:
Hello! Sorry to reopen this one, but typescript still complains about the `ref` attribute.

The error is:

```
Property 'ref' does not exist on type 'IntrinsicAttributes &...
```

I'm using the `9.0.0-alpha.4` version, you can check it out in [this codesandbox](https://codesandbox.io/s/stitchesstyled-ref-rv0m9?file=/src/App.tsx)

hadihallak:
@rubenmoya sorry about that. the fix above was for the library itself not forwarding refs internally and not the type, however, this issue along with a few other type issues have already been resolved and are scheduled for a release on Monday 😄

hadihallak:
Fixed in #134

Issue #96, opened 2020-08-14
TS: expects `children: never` when using `styled(Component, {})`  


jjenzz:
https://codesandbox.io/s/objective-sunset-3rpyp

The primitives have their own type so we cannot style any of our primitives using this API atm.

![image](https://user-images.githubusercontent.com/175330/90255919-db7c9380-de3c-11ea-8aff-5566194e8cc1.png)

![image](https://user-images.githubusercontent.com/175330/90255940-e33c3800-de3c-11ea-8cfb-c7b2f01efa5e.png)

peduarte:
Without React.FC we get a different error https://codesandbox.io/s/recursing-http-ge42v?file=/src/App.tsx

christianalfoni:
I believe this was fixed with: https://github.com/modulz/stitches/commit/842a55bb0447cd01c08dd6f267477924fd74f66d as well

hadihallak:
@christianalfoni 👋  just an FYI that most type issues are already fixed as part of #134 which I will be merging in the next hours

hadihallak:
Fixed in #134

Issue #85, opened 2020-08-11
Update the variant syntax


colmtuite:
After chatting with @peduarte, we agreed the variant syntax could be improved. After testing the lib, I think it gets very difficult to scan when you have lots of variants.

We agreed on this syntax:

```
const Button = styled(element, {
  appearance: 'none'
},
{
  variant: {
    // styles
  }
},
{
  size: {
    // styles
  }
})
```

However, myself and @StephenHaney still think totally separate variant functions is the best option.

peduarte:
Yep, we could support this approach. It's similar to how Styled Components function can receive many comma-separated style objects.

Should we continue to support multiple variants in one object? I don't see any harm in doing so.

christianalfoni:
I am waiting with this for a decision. I totally see the "scanning argument". I would argue it feels a bit unnatural to split up into multiple objects where you are only allowed to have one key. But if allowing multiple keys, the reason for multiple arguments looses its purpose.

So this is like a syntax discussion, which is scary in itself 😂 But you can improve scanning by adding quotes for example:

```ts
const Button = styled(element, {
  appearance: 'none'
},
{
  "variant": {
    // styles
  },
  "size": {
    // styles
  }
}
```

Or make the values constants:

```ts
const Button = styled(element, {
  appearance: 'none'
},
{
  variant: {
    MUTED: {}
  },
  size: {
    SMALL: {},
    LARGE: {}
  }
}
```

It seems that the multiple arguments is a compromise? How does the separate variants function look like? 😄

peduarte:
@christianalfoni we had considered separate functions as well, something like:
```jsx
const Button = styled();

Button.variant('size', {
  small: {},
  large: {}
})
```

But im worried about increasing too much the API surface area. Maybe not a problem. Whats your opinion on this? I know that Colm likes it.

colmtuite:
@peduarte in your recent tweet I saw you had line breaks between the variants. Even that helps a lot I think. So maybe it is just a simple syntax preference thing, not sure.

I think @StephenHaney likes the separate functions approach too?

StephenHaney:
I do, all else being equal, but I do acknowledge it's likely more complicated to implement. If we do go with a function call per variant, I like Pedro's idea of supporting both syntaxes as well.

jjenzz:
> Also, a bit weird to attach functions to a React Component?

I think this is okay tbh. With a functional API, consumers could also do:

```jsx
const Button = styled('button', {

[...]

PR #134, opened 2020-08-26
New variants api and type fixes

hadihallak:
**What's inside:**
- ~Updated variants api according to #85~
We went with a new syntax since the syntax suggested in #85 can't be typed unless chained:
new variant syntax:
```
  // Base styles:
  backgroundColor: "red",
  // Variants are scoped into a variants property:
  variants: {
    variant: {
      red: {},
      black: {},
      blue: {},
      yellow: {},
    },
    size: {
      1: {},
      2: {},
      3: {},
      4: {},
    },
  },
});
```
- Compound variants #80 
- Fix forward ref types #71 
- Fix: expects `children: never` when using `styled(Component, {})` #96
- Fix variant values not working when non-string value is provided #47
- Fix refs throwing type errors #110
- Fix value autocomplete isn't working inside variants. #133
- Fixs gigantic type file size when building a ds
- Improve variants types in props so now when you create a variant like with the name true, it will automatically create the corresponding boolean variant in addition to the string one so:
```
  variants: {
    isActive: {
      true: {},
      false: {},
    }
  }
})

// now you will be able to use it like so
<Button isActive={true} />
or
<Button isActive="true" />
```
with both of the options being strictly typed to the actual values

**Please hammer this down with testing for both typing and actual functionality as there was a fair bit of types/functionality was re-written and we currently don't have testing setup for the styled package so play around with it and let me know if you find any bugs**


peduarte:
@christianalfoni would you mind having a look as well, pleeeeease? 🙏

peduarte:
regarding the variant API, thanks for that, im looking at the codesandbox and doing some tests

So to clarify, in order for types to work properly, the variant calls need to be chained.
```
const Button = styled('button', {})
  .variant('color', {})
  .variant('size', {})
```

Instead of the original idea, which was to have completely seperate blocks:
```
const Button = styled('button', {})
Button.variant('color', {})
Button.variant('size', {})
```

Personally, Im not a fan of chaining, but I could live with it. What worries me the most is the confusion around the fact that types will not work unless you chain. These things become really hard to document, and will drive people nuts. Because from a technical perspective, you’d think that if you can chain, you can also call Button.variant - which you can, but… types wont work

peduarte:
typing stuff seems to be working very well though

PR #140, opened 2020-08-28
Add global styles support

### LINKED ISSUE #65: Add a `createGlobalStyle` function
We need to be able to create global function, similar to Styled Component. This should also support tokens and screens, but not variants.

hadihallak:
Resolves #65


packages/css/tests/index.test.ts
```diff
@@ -1183,4 +1183,61 @@ describe("createCss", () => {
...
+      @media (min-width: 700px){ div{color:red;}}"
+    `);
+  });
+    test("should error when styles are used without nesting", () => {
+    const css = createCss({}, null);
+    expect(()=> {
+      css.global({
+          background: 'red'
+        })
+    }).toThrow()
```
peduarte, line 1240: Im not following this use case. Why do they need to be nested?

To be clear, we need to be able to [do this with the global CSS](https://github.com/modulz/radix/blob/master/packages/radix/src/GlobalStyles.tsx#L4-L19) function

The main reason for this is to remove styles from `body`, `html` and have access to tokens and breakpoints

peduarte (commented):
Question about nesting 😄

packages/css/tests/index.test.ts
hadihallak: Because not having nesting at all doesn't make sense unless you're creating a class to attach the styles to and in this case the styles wouldn't be global (if they're not inside an explicit selector like body or HTML etc..)

```
   css.global({
          background: 'red'
        })
```

this code above can't be global since it lacks a selector and should error ^

peduarte: @hadihallak ahhh yeah, of course 😓 

I was thrown off by the wording. Can we be more explicit about the need of defining a selector?

PR #152, opened 2020-09-02
Add spec font shorthand parsing

jonathantneal:
Sweet project! I’ve just started looking around, and you made it easy to find one’s way around. ✨ 

This PR adds spec-compliant parsing of the [CSS shorthand `font` property](https://drafts.csswg.org/css-fonts-4/#font-prop) as well as some tests to ensure this is so. This change may be considered substantial, and so I’d love to work with you on ensuring it upholds the standards and “_voice_” of your project.

This change represents a parsing improvement, and it avoids changing the behavior of shorthands as they are in this project. In CSS, the `font` shorthand resets all subproperties of the font property to their initial values, while in this project the `font` shorthand appears additive.

The tests in this PR are taken from example 18 of the [CSS Fonts Module Level 4 specification](https://drafts.csswg.org/css-fonts-4/#example-43f2026b), which are as follows:

```css
p { font: 12pt/14pt sans-serif }
p { font: 80% sans-serif }
p { font: x-large/110% "new century schoolbook", serif }
p { font: bold italic large Palatino, serif }
p { font: normal small-caps 120%/120% fantasy }
p { font: condensed oblique 12pt "Helvetica Neue", serif; }
p { font: condensed oblique 25deg 753 12pt "Helvetica Neue", serif; }
```

I’ll leave a few notes in the code to highlight certain choices I made. Thank you for considering this PR. It was a lot of fun to make. I picked the `font` shorthand because I had always wondered how the browser went about parsing these. 😄


packages/css/src/shorthand-parser/index.ts
```diff
@@ -4,6 +4,10 @@ import { tokenizeValue } from "./value-tokenizer";
 const unitMatch = /^[0-9.]+[a-z|%]/;
 const easingMatch = /\(.*\)|ease|ease-in|ease-out|ease-in-out|linear|step-start|step-end/;
 
+const fontSizeMatch = /^([+-]?[\d.]+([a-z]+|%)?|large(r)?|medium|small(er)?|x{1,3}-large|x{1,2}-small)(\/[+-]?[\d.]+([a-z]+|%)?)?$/;
+const fontStyleMatch = /^[+-]?[\d.]+deg$/;
+const fontWeightMatch = /^(0*[1-9]\d{0,2}|1000|bold(er)?|lighter)$/;
```
jonathantneal, line 9: I only now notice that this project is using `[0-9]`, while I am using `\d`. Let me know if I should adjust these. These regular expressions are otherwise intended to follow the patterns seen in `unitMatch` and `easingMatch`.

Note: A more accurate regular expression for CSS numbers may be `/([+-]?(?:\d+(?:\.\d*)?|\.\d+)(?:[Ee][+-]?\d+)?)/`. However, if parsing performance is of importance, it may be better to identify numbers at the tokenization stage and leave out regular expressions altogether.

packages/css/src/shorthand-parser/index.ts
```diff
@@ -129,31 +133,85 @@ export const animation = createPropertyParser(
...
+
+          break;
+
+        case fontWeightMatch.test(lower):
+          css.fontWeight = tokens.fontWeights[value] || value;
+
+          break;
+
+        case fontSizeMatch.test(lower):
+          chains.shouldParseFontFamily = true;
```
jonathantneal, line 165: For this, I may deserve the bell of shame. 🔔😰 This is me, a new contributor, desperate to avoid modifying `createPropertyParser`, resorting to adding named keys to the `chains` array.   This is done to ensure that all property parts after the font size are interpreted as font families. ([see usage](https://github.com/modulz/stitches/pull/152/files#diff-0ec06497f9094ba10b71f75d395050c6R145))

packages/css/src/shorthand-parser/index.ts
```diff
@@ -4,6 +4,10 @@ import { tokenizeValue } from "./value-tokenizer";
 const unitMatch = /^[0-9.]+[a-z|%]/;
 const easingMatch = /\(.*\)|ease|ease-in|ease-out|ease-in-out|linear|step-start|step-end/;
 
+const fontSizeMatch = /^([+-]?[0-9.]+([a-z]+|%)?|large(r)?|medium|small(er)?|x{1,3}-large|x{1,2}-small)(\/[+-]?[0-9.]+([a-z]+|%)?)?$/;
+const fontStyleMatch = /^[+-]?[0-9.]+deg$/;
+const fontWeightMatch = /^(0*[1-9][0-9]{0,2}|1000|bold(er)?|lighter)$/;
```
jonathantneal, line 9: These regular expressions are intended to follow the patterns seen in `unitMatch` and `easingMatch`.

A more accurate regular expression for CSS numbers may be `/([+-]?(?:\d+(?:\.\d*)?|\.\d+)(?:[Ee][+-]?\d+)?)/`. However, if parsing performance is of importance, it may be _even better_ to identify numbers at the tokenization stage and leave out these regular expressions altogether.

packages/css/src/shorthand-parser/index.ts
```diff
@@ -129,31 +133,84 @@ export const animation = createPropertyParser(
...
+          css.fontSize = tokens.fontSizes[fontSize] || fontSize;
+
+          if (lineHeight) {
+            css.lineHeight = tokens.lineHeights[lineHeight] || lineHeight;
+          }
+
+          break;
+
+        case lower === "italic":
+        case lower === "oblique":
```
jonathantneal, line 177: These string equality checks are used because they seem more clear and are likely much faster than regular expressions checks.

packages/css/tests/shorthand-parser.test.ts
```diff
@@ -151,7 +151,7 @@ describe("Font shorthand", () => {
         "fontFamily": "Georgia,serif",
         "fontSize": "12px",
         "fontStyle": "italic",
-        "fontWeight": 700,
+        "fontWeight": "bold",
```
jonathantneal, line 154: A relatively harmless but unintended side effect of this change is that the `bold` keyword no longer transforms into `700`. However, it’s worth noting that in the previous version, a weight like `bolder` would have inaccurately become `700`.

peduarte:
Thanks for this, looks like a great addition! Good catch on the `bold` keyword value!

@hadihallak and @christianalfoni if you guys can review this would be 👌

packages/css/src/shorthand-parser/index.ts
hadihallak: Yeah I totally agree with that 👍 
the tokenizer used to support this functionality, it would return an array of tokens with types belonging to 
```string | number | number with unit | CSS function```
but we weren't sure at the time how we were going to use it so we kept it as simple as possible but now i'm thinking we should bring this functionality back.

we can change this when we update the tokenizer but for now, this is good

**Review by hadihallak — COMMENTED** [review]
Thank you so much @jonathantneal for submitting this PR. looks amazing 💯 

>the font shorthand resets all subproperties of the font property to their initial values, while in this project the font shorthand appears additive.

yeah, that's correct. this is also true for almost all shorthand. 
we're likely to change this to include initial values once we we iron out the library internally but as you can see this is a large project on its own if we want to get near-perfect correctness for shorthand and this parsing accuracy will result in a larger library size so the plan is to have a plugins system in place for stitches which will allow us to move the shorthand parsing outside the library so that the extra size required to support shorthands is optional.

Again, thank you so much for contributing 💪 
would you mind resolving the merge conflicts so that we can merge? or if you don't mind I could resolve the conflicts and merge. whichever you prefer.

**Comment by jonathantneal** [comment]
> would you mind resolving the merge conflicts so that we can merge? or if you don't mind I could resolve the conflicts and merge. whichever you prefer.

Thanks again for reviewing my PR, @hadihallak! I’ve rebased with master and committed the rebase changes from your review in cf827fa.

File: packages/core/src/shorthand-parser/index.ts [hunk 6]
```diff
@@ -92,26 +96,77 @@ export const animation = createPropertyParser((_: any, css: any, value: any, ind
...
+        case lower === 'small-caption':
+        case lower === 'status-bar':
+          chains.shouldParseFontFamily = true;
+
+          css.fontFamily = setChainedValue(css.fontFamily, tokens.fonts[value] || value);
+
+          break;
+
+        default:
+          break;
```
jonathantneal, line 165: This was added during the rebase in order to respect an eslint/tslint rule which I had not encountered beforehand. An alternative that I considered was `// eslint-disable-next-line switch-default` over the initial `switch` statement, a technique I saw used elsewhere. However, the rule _exists_, and so it seemed better to include the `default` statement to be explicit (tho “noisey” to me), rather than immediately suggesting a change your linting configuration.

hadihallak: Sounds good. we are going to be reviewing the lining rules at a later time but for now, it's a good idea to stitch with them.

PR #158, opened 2020-09-02
[WIP] Vue 3 support

baryla:
Great work on the lib everyone! 

I've actually built something very similar (at least from the API perspective) for a Vue 3 project I've been working on but never released and when I saw this project this morning on Twitter ... exactly what I need haha! 

So here's a WIP PR for Vue 3 and it fully works. There are some typing issues (which I will fix), duplicated code from the React implementation (which can moved to `core/utils`?), some minor bits which need to be rewritten as well as missing tests but I just wanted to know if this is something you guys would be happy to merge once all of the above are sorted out? I'm quite happy tackling them all :) 

I saw that Vue was something you guys considered but via a directive which I don't think is the best route to go so this implementation allows for creating styled components exactly like in React.


baryla:
Just a small update - I haven't been able to dedicate the time I wanted to this PR but I should be able to now.

baryla:
The `@stitches/vue` package should now on the same level as the `@stitches/react` one in terms of types and functionality. 

One thing which we can't support for `@stitches/vue` without modifying `@stitches/core` is the ability to set the theme on the `class` attribute so therefore, we have to use `className`. This isn't ideal for Vue devs because even in JSX, `class` attribute is what is being used in favour of `className`. The problem here: https://github.com/modulz/stitches/blob/canary/packages/core/src/index.ts#L589. In Vue, the `class` property uses keys on an object as classes therefore `toString` is completely ignored and we end up with an element such as:
```html
<button class="name definition toString">Hello World</button>
```
---

Obviously there are many duplicates in this implementation compared to the react one so maybe it's worth moving most of the implementation to a shared file? Almost everything can be shared across the two frameworks. 

Would love to see your thoughts on this.

aarongarciah:
This is awesome! Do you have some sandbox with a working example?

baryla:
@aarongarciah here you go: https://codesandbox.io/s/stitches-vue-3-jxyfs?file=/src/App.vue.

It's not the best example as it's missing the types but at least you can see it working with Vue 3. Let me know if you find any bugs :)

leonardotessaroalves:
Any plans to have Vue 2 integration?

package.json
```diff
@@ -41,7 +42,8 @@
     "@babel/preset-env": "^7.11.5",
     "@babel/preset-typescript": "^7.10.4",
     "@preconstruct/cli": "^1.1.27",
-    "react": "^16.13.1"
+    "react": "^16.13.1",
+    "vue": "3.0.0-rc.10"
```
eshimischi, line 46: Vue 3.0 is released already https://github.com/vuejs/vue-next/releases

baryla:
Another update - I spoke to @peduarte around 2 weeks ago and we decided that we won't merge this PR anytime soon until the API is final for a v1 for `@stitches/core`. I will however, every couple of days/weeks update this PR to align the Vue package with the React package so that in the future, once v1 is nearing, it won't be a massive job. 

@leonardotessaroalves - not at the moment and I'm not sure that we should. Depending on how stitches evolves, we would end up having to use something like `vue-demi` to align Vue 2 and 3 API's.

package.json
baryla: Good spot buddy. I'll update this when I get some free time :)

peduarte:
@baryla we're in the process of refactoring the core codebase. Its no longer compatible with this PR so I'll close it.

But, the good news is, we're making it much easier to support other libraries like Vue! I'll let you know when it's done. 

Thanks so much <3

baryla:
@peduarte nice one buddy! That's excellent news, looking forward to it :)

Issue #142, opened 2020-08-31
Parse `text-decoration` shorthand


colmtuite:
The `text-decoration` longhand properties work in browsers these days, so we need to parse this with color tokens.

vivek12345:
@peduarte Can I look into this with obviously some help from you? 
Note: Contributing for the first time to a css in js library.

peduarte:
Sure. Make sure to add @hadihallak to the review and ping him if you need anything 👍

vivek12345:
Thank you. @hadihallak I have done some initial work around text-decoration shorthand and long hand properties. Would be super helpful if I can get some feedback around it and improve it to make it ready.

hadihallak:
@vivek12345 Thank you so much for the contribution. will be taking a look in a couple of hours.

PR #159, opened 2020-09-02
feat: added support for text decoration shorthand and other long hand properties

vivek12345:
* Added support for text-decoration shorthand. Now we can pass in values like
```
textDecoration: underline dotted red;
```
or
```
 textDecoration: underline overline dotted gray400;
```

* Also added support for text-decoration-thickness in short hand properties 
```
 textDecoration: underline overline dotted gray400 5px;
```

* Tests updated for all of the above changes

Using this doc as a reference
https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration

Fix https://github.com/modulz/stitches/issues/142

### DIFF
--- a/packages/core/src/shorthand-parser/index.ts
+++ b/packages/core/src/shorthand-parser/index.ts
@@ -299,13 +299,13 @@ export const textDecoration = createPropertyParser((tokens: any, css: any, value
 });
 
 export const textDecorationStyle = createPropertyParser((tokens: any, css: any, value: any) => {
-  if (value.match(/solid|double|dotted|dashed|wavy/)) {
+  if (value.match(/solid|double|dotted|dashed|wavy|unset/)) {
     css.textDecorationStyle = value;
   }
 });
 
 export const textDecorationLine = createPropertyParser((tokens: any, css: any, value: any) => {
-  if (value.match(/none|underline|overline|line-through|blink/)) {
+  if (value.match(/none|underline|overline|line-through|blink|unset/)) {
     css.textDecorationLine = setChainedValue(css.textDecorationLine, value, ' ');
   }
 });
--- a/packages/core/src/shorthand-parser/index.ts
+++ b/packages/core/src/shorthand-parser/index.ts
@@ -296,43 +296,17 @@ export const boxShadow = (tokens: any, value: string) => {
 
 export const textDecoration = createPropertyParser((tokens: any, css: any, value: any) => {
   if (matchString(value, /unset/)) {
-    // we are checking the existing value here because we want to retain the long hand property that was previously applied
-    /**
-     * For example:
-     * textDecorationColor: red
-     * textDecoration: unset
-     *
-     * The above should result in a final css rule like the following
-     * text-decoration-style: unset;
-     * text-decoration-line: unset;
-     * text-decoration-color: red
-     *
-     * This is as per the css rules
-     */
-    css.textDecorationStyle = css.textDecorationStyle || value;
-    css.textDecorationLine = css.textDecorationLine || value;
-    css.textDecorationColor = css.textDecorationColor || value;
+    css.textDecorationStyle = value;
+    css.textDecorationLine = value;
+    css.textDecorationColor = value;
+    css.textDecorationThickness = value;
   } else if (matchString(value, /solid|double|dotted|dashed|wavy/)) {
     css.textDecorationStyle = value;
-  } else if (value.match(/none|underline|overline|line-through|blink/)) {
+  } else if (matchString(value, /none|underline|overline|line-through|blink/)) {
     css.textDecorationLine = setChainedValue(css.textDecorationLine, value, ' ');
+  } else if (matchString(value, unitMatch) || matchString(value, /auto|from-font/)) {
+    css.textDecorationThickness = value;
   } else {
     css.textDecorationColor = tokens.colors[value] || value;
   }
-});
-
-export const textDecorationStyle = createPropertyParser((tokens: any, css: any, value: any) => {
-  if (matchString(value, /solid|double|dotted|dashed|wavy|unset/)) {
-    css.textDecorationStyle = value;
-  }
-});
-
-export const textDecorationLine = createPropertyParser((tokens: any, css: any, value: any) => {
-  if (matchString(value, /none|underline|overline|line-through|blink|unset/)) {
-    css.textDecorationLine = setChainedValue(css.textDecorationLine, value, ' ');
-  }
-});
-
-export const textDecorationColor = createPropertyParser((tokens: any, css: any, value: any) => {
-  css.textDecorationColor = tokens.colors[value] || value;
 });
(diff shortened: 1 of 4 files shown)


hadihallak:
Hey, @vivek12345  looks great! let me know when this is done and I will take another look.

Thanks for contributing 🚀

vivek12345:
## For longhand notation
I have added property value `unset` for long-hand notations.
So with the current implementation we allow
```
text-decoration-line: unset 
```
and
```
text-decoration-style: unset 
```
and
```
text-decoration-color: unset 
```

## For shorthand notation
if the user enters value
```
textDecoration: unset
```
we will create the following rule
```
text-decoration-line: unset 
text-decoration-style: unset 
text-decoration-color: unset 
```

vivek12345:
@hadihallak I have a question around two property values: `initial` and `inherit`. How do we handle that?

Also should we also throw an error if someone enters a css value for text-decoration which is not allowed
For example:
```
text-decoration:  unset red;
```
this as per chrome inspector is not allowed but as per our parsing we will allow this.

What do you suggest we should do for the above case?

vivek12345:
I think I can make use of the chain property to detect an unsupported value like 
```
textDecoration: unset red;
```
and then throw an error.

Let me know your thoughts on this?

packages/core/src/shorthand-parser/index.ts
```diff
@@ -292,3 +293,46 @@ export const boxShadow = (tokens: any, value: string) => {
...
+    /**
+     * For example:
+     * textDecorationColor: red
+     * textDecoration: unset
+     *
+     * The above should result in a final css rule like the following
+     * text-decoration-style: unset;
+     * text-decoration-line: unset;
+     * text-decoration-color: red
+     *
```
hadihallak, line 309: Could you please walk me through this change as the css argument is an empty object that we append parsed longhand values onto so I'm not sure how this condition would be met. 

Would appreciate if you could write a test case for it as it would make it easier to understand

packages/core/src/shorthand-parser/index.ts
```diff
@@ -292,3 +293,46 @@ export const boxShadow = (tokens: any, value: string) => {
...
+     */
+    css.textDecorationStyle = css.textDecorationStyle || value;
+    css.textDecorationLine = css.textDecorationLine || value;
+    css.textDecorationColor = css.textDecorationColor || value;
+  } else if (matchString(value, /solid|double|dotted|dashed|wavy/)) {
+    css.textDecorationStyle = value;
+  } else if (value.match(/none|underline|overline|line-through|blink/)) {
+    css.textDecorationLine = setChainedValue(css.textDecorationLine, value, ' ');
+  } else {
+    css.textDecorationColor = tokens.colors[value] || value;
```
hadihallak, line 320: We need to handle `text-decoration-thickness` in here as well and then add a test case that includes parsing it

packages/core/src/shorthand-parser/index.ts
```diff
@@ -292,3 +293,46 @@ export const boxShadow = (tokens: any, value: string) => {
...
+
+export const textDecorationLine = createPropertyParser((tokens: any, css: any, value: any) => {
+  if (matchString(value, /none|underline|overline|line-through|blink|unset/)) {
+    css.textDecorationLine = setChainedValue(css.textDecorationLine, value, ' ');
+  }
+});
+
+export const textDecorationColor = createPropertyParser((tokens: any, css: any, value: any) => {
+  css.textDecorationColor = tokens.colors[value] || value;
+});
```
hadihallak, line 338: We can remove these as in addition to all related tests as they're already longhand syntax so parsing them isn't doing anything as you can see in their tests

hadihallak (commented):
Hey @vivek12345 👋 
Amazing first PR. love how you kept everything in line with the code style in the codebase 💯 

I left a few comments that I would appreciate if you could resolve before we can merge this in.
let me know if you have any questions

hadihallak:
> I have a question around two property values: initial and inherit. How do we handle that?

I don't think we're currently doing any specific handling for these properties. let me know if you have any ideas for them

> Also should we also throw an error if someone enters a css value for text-decoration which is not allowed
For example:

for now, we're trying to get as close as possible to the css spec for **valid** values. having errors is nice to have but not required at this stage.
we will likely expand this to include detailed error messages at a later time when we move the shorthand parsing to its own package but for now, the focus is on valid values.

> I think I can make use of the chain property to detect an unsupported value like

> textDecoration: unset red;
as the chain is usually an array of all these values so it would be ['unset', 'red'] and then I can run some check to throw an error.

sounds good but as I mentioned, not erroring for invalid values is fine for this stage

packages/core/src/shorthand-parser/index.ts
vivek12345: @hadihallak I did not understand this. If I remove these how will this syntax be parsed?

packages/core/src/shorthand-parser/index.ts
vivek12345: text-decoration-thickness seem to be only supported on firefox and safari at the moment. Do we still need to add it?
https://caniuse.com/?search=text-decoration-thickness

packages/core/src/shorthand-parser/index.ts
vivek12345: Actually I did understand now. Thank you for pointing it out 👍

packages/core/src/shorthand-parser/index.ts
vivek12345: You are right. I incorrectly assumed that the previous css property will exist on the css object.
So what I was trying to achieve by doing this was to mimic the browser behaviour.
On chrome when I was trying to do the following
```
text-decoration-color: red;
text-decoration: unset;
```

It was actually resulting in the following css:
```
text-decoration-color: red;
text-decoration-line: unset;
text-decoration-style: unset;
```

But since our css object is always initialized as an empty object, it seems like we can't do this.

vivek12345:
* Added support for text-decoration-thickness
* Removed code for text-decoration long hand properties along with the corresponding tests
* Added more tests for verifying the token usage

All of this can be tested at the sandbox here
https://codesandbox.io/s/react-forked-f54ei?file=/src/App.js

vivek12345:
@hadihallak This is ready now with all the changes.

peduarte:
@vivek12345 abdul is on holiday atm but I'll catch up with this later today or tomorrow. Thanks

peduarte:
LGTM.

Gonna get this in `canary` now.

BTW how did you do this in CodeSandbox:
```
"@stitches/core": "https://pkg.csb.dev/vivek12345/stitches/commit/be733f97/@stitches/core",
"@stitches/react": "https://pkg.csb.dev/vivek12345/stitches/commit/be733f97/@stitches/react"
```

vivek12345:
Thank you @peduarte . I actually added codesandbox ci on the fork of the repo and created the same pr on my repo as well.
codesandbox ci builts every PR and deploys it on that `pkg.csb.dev` domain.

You can check the config here
https://github.com/vivek12345/stitches/blob/master/.codesandbox/ci.json

We can do this for modulz/stitches repo too but the owner first needs to add the codesandbox ci and then I can configure the rest.

Let me know if we need to do this.
https://codesandbox.io/docs/ci

peduarte:
@vivek12345 yeah that'd be good. Abdul raised an issue about this actually, #163 

Let me take a look

peduarte:
Done in #198 

Thanks!

PR #161, opened 2020-09-03
Bug/breakpoints media queries precedence

### LINKED ISSUE #157: @supports in @media results in incorrect rule specificity
I was trying out the `utils` functionality and combined it with a media query and noticed that some rules were not being overridden by the supports declarations, e.g:

```jsx
import { createStyled } from '@stitches/react';

export const { styled, css } = createStyled({
  breakpoints: {
    large() {
      return `@media (min-width: 64em) { ${rule} }`;
    }
  },
  utils: {
    maxLines() {
      return (value) => {
        return {
          whiteSpace: 'nowrap',
          overflow: 'hidden',
          textOverflow: 'ellipsis',

          '@supports (-webkit-box-orient: vertical)': {
            display: '-webkit-box',
            whiteSpace: 'normal',
            overflow: 'none',
            WebkitLineClamp: value,
            WebkitBoxOrient: 'vertical',
          },
        };
      };
    },
  },
}

export const Text = styled('p', {});

/**
 * This will not work because some styles such as `overflow` and `whiteSpace`
 * from the @supports are too low in specificity, the first set of rules seem
 * to always win.
 */
<Text css={{
  maxLines: 1,

  large: {
    maxLines: 3, // :sadpepe:
  }
}} />
```

[A reduced example can be found here](https://codesandbox.io/s/stitches-util-ideas-forked-ttdgs) where I set the color of the text to blue and if a `@supports` rule passes the color is set to red. The media query used is set to a min-width of 1em so you should see the red text immediately. Inspecting through dev tools show it's being set but the specificity is too low, resulting in blue text.

hadihallak:
Fixes #157

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -250,7 +250,11 @@ const createToString = (
     const shouldInject = !preInjectedRules.size || !preInjectedRules.has(`.${className}`);
 
     if (shouldInject) {
-      sheets[this.breakpoint].insertRule(createCssRule(breakpoints, this, className));
+      const sheet = sheets[this.breakpoint];
+      sheet.insertRule(
+        createCssRule(breakpoints, this, className),
+        this.inlineMediaQueries.length ? sheet.cssRules.length : 0
+      );
     }
 
     // We are switching this atom from IAtom simpler representation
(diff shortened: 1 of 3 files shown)


peduarte (approved):
👌

swudged:
🙌

Issue #175, opened 2020-09-04
Transition tokens not working?


imjoshellis:
## Describe the bug

Not if I'm missing something about the syntax, but I can't seem to get tokens working with transitions. 

## To Reproduce

minimal example: https://codesandbox.io/s/keen-fermat-4jnoz?file=/src/App.js

1. Setup stitches and stitches.config.js
2. Add transition token `$slow: "all 1000ms"`
3. Add `$slow` token as value for `transition` in `styled()`
4. Hover over the component 

Notes: 
- Setting the exact same transition "inline" within the `styled()` object like `transition: "all 1000ms"` works as expected.
- Color tokens are working as expected.

## Expected

"Inline Transition" and "Token Transition" in the example to have identical behavior.

peduarte:
Looks like a bug 🐛 

Thanks for raising it!

imjoshellis:
I also noticed after submitting this that `transitions` isn't listed in the property mapping table at the bottom of the  [token page of the docs](https://stitches.dev/docs/tokens#property-mapping).

vivek12345:
I can take a look at fixing this

hadihallak:
@vivek12345 That's awesome! looking forward to it 💯 
let me know if you need help with anything.

vivek12345:
Sure I will 👍

peduarte:
We've published a pre-release of Stitches beta 🚀

**Core**:
```bash
npm install @stitches/core@0.1.0-canary.0
```

**React**:
```bash
npm install @stitches/react@0.1.0-canary.0
```

This issue should have been resolved in the beta version. We'd appreciate it if you could verify it. 

⚠️ Please refer to this [migration guide](https://stitches-site-git-beta.modulz-deploys.com/blog/migrating-from-alpha-to-beta) as there are some breaking changes.

PR #181, opened 2020-09-05
bugfix: check if the value passed to transition is a token and then use the token value

vivek12345:
**What is the Issue?**
https://github.com/modulz/stitches/issues/175

**What kind of change does this PR introduce?**
This PR is a bug fix.

**What is the current behavior?**
In the current behavior, we do not set transition property values if a token is passed instead of a value.

**What is the new behavior?**
In the new behavior, we will first check if the value passed is a token and if it is then we will use the token value to set properties of transition.
In case it's a normal value, we will continue to use that value.

**Checklist**:

- [x] Tests
- [x] Ready to be merged 
- [x] Issue Linked?

### DIFF
--- a/packages/core/src/shorthand-parser/index.ts
+++ b/packages/core/src/shorthand-parser/index.ts
@@ -168,22 +168,23 @@ export const font = createPropertyParser(
   }
 );
 
-export const transition = createPropertyParser(
+export const transition = (tokens: any, value: string) => {
   // The whole token is a transition, so need to grab it before passing in here
-  (_: any, css: any, value: any, index: any, chain: any) => {
-    if (matchString(value, unitMatch)) {
+  const tokenOrValue = tokens.transitions[value] || value;
+  return createPropertyParser((_: any, css: any, parsedValue: any, index: any, chain: any) => {
+    if (matchString(parsedValue, unitMatch)) {
       if (chain.findIndex((part: any) => part.match(unitMatch)) === index) {
-        css.transitionDuration = setChainedValue(css.transitionDuration, value);
+        css.transitionDuration = setChainedValue(css.transitionDuration, parsedValue);
       } else {
-        css.transitionDelay = setChainedValue(css.transitionDelay, value);
+        css.transitionDelay = setChainedValue(css.transitionDelay, parsedValue);
       }
-    } else if (matchString(value, easingMatch)) {
-      css.transitionTimingFunction = setChainedValue(css.transitionTimingFunction, value);
+    } else if (matchString(parsedValue, easingMatch)) {
+      css.transitionTimingFunction = setChainedValue(css.transitionTimingFunction, parsedValue);
     } else {
-      css.transitionProperty = setChainedValue(css.transitionProperty, value);
+      css.transitionProperty = setChainedValue(css.transitionProperty, parsedValue);
     }
-  }
-);
+  })(tokens, tokenOrValue);
+};
 
 export const margin = createPropertyParser((tokens: any, css: any, value: any, index: any) => {
   if (index === 0) {
--- a/packages/core/tests/shorthand-parser.test.ts
+++ b/packages/core/tests/shorthand-parser.test.ts
@@ -28,6 +28,7 @@ const tokens = {
   fonts: { main: 'potato-font' },
   borderWidths: { hairLine: '1px' },
   radii: { 1: '3px' },
+  transitions: { slow: 'all 1000ms', multiGroup: 'margin-right 2s, color 1s' },
 };
 
 // works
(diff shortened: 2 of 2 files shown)


vivek12345:
@hadihallak I need some help here in understanding how to deal with this scenario
```
tokens: {
  transitions: {
     slow: 'all 1000ms'
  }
}
```
This gets converted into a css variable like the following:
```
transitions: { slow: 'var(--transitions-slow)' }
```

Now when it reaches our shorthand parser, how do I get the value of transition-property and transition-duration from it?

Right now If I tweak the regex of easingMatch to
```
const easingMatch = /(cubic-bezier|steps)\(.*\)|ease|ease-in|ease-out|ease-in-out|linear|step-start|step-end/;
```

It gets's assigned to `transitionProperty: var(--transitions-slow)`

What I actually need is 
```
transitionProperty: all
transitionDuration: 1000ms
```

hadihallak:
@vivek12345
Good catch. yeah, this is making me wonder if having the whole transition rule as a token is a good idea.

Here is what I'm currently thinking:
**A. we rework how transition tokens are stored in the tokens object so that they reflect longhand properties**

**B. Instead of generating a token for the whole transition value, we generate tokens for the parsed shorthand rule so for this:**
``` 
tokens: {
  transitions: {
     slow: 'all 1000ms'
  }
}
```
we would internally parse it into two tokens:
```
  --transitions-slow-property: 'all'
  --transitions-slow-duration: '1000ms'
```
so that then when the tokens are used we are able to infer the name from the value so if the token value was `slow` we would check if the user has a token with that name in their tokens so `tokens.transitions[value]` and if we found that the user defined a value there we would infer the preparsed tokens so if we want the duration token we would do
`--transitions-${value}-duration`

I think option B would provide the best experience but I'm also keen to know your opinion. 

@peduarte what do you think?

imjoshellis:
My $0.02 - I'd prefer to set transition tokens longhand but combine them shorthand in the styles object: `transition: all $slow $customEasing`

This way you can have a few duration tokens and a few custom easing tokens without having to maintain tokens of every combo.

Not sure how feasible that is, though. I know transition property is kinda weird, and I haven't dug into the code.

vivek12345:
Thank you for your inputs @hadihallak . After thinking about both the approaches
A) Seems easier for us but then the users would have to type all the tokens for long hand values
B) We can do this and I like this approach too, although I do have a question around it

In the second case will we also change our tokens to something like
```
transitions: {
    'slow-property: 'var(--transitions-slow-property)',
    'slow-duration': 'var(--transitions-slow-duration)'
},
```
If we do not do this then our transition token will continue to have the following structure
```
transitions: {
    slow: "all 1000ms"
},
```

In the above case then we would actually not be using css vars.

We can discuss if I have understood this incorrectly.

Also why do convert these values into vars? shorthand parser is anyway going to parse them then why add a css variable?

vivek12345:
After giving this another thought as suggested by @hadihallak , here is one approach that I think we can try
When someone defines a transition token like the following
```
tokens: {
  transitions: {
     slow: 'all 1000ms'
  }
}
```

We will create two css variables for this
```
var(--transitions-slow-property)
var(--transitions-slow-duration)
```

Internally we will convert the tokens object into something like this
```
tokens: {
  transitions: {
     slow: {
       original: "all 1000ms",
       property: "all",
       duration: "1000ms"
     }
  }
}
```

With this,  in shorthand parser we will first check if a value is a token. So if someone passes value as "slow"

We will check `tokens.transitions[value]` and get an object consisting of the following structure
```
{
   original: "all 1000ms",
   property:  var(--transitions-slow-property),
   duration:  var(--transitions-slow-duration)
}
```

So now if a value exists in our token object we would parse the original value through chain parser and then when we are setting the value we will set the property like this:
```
css.transitionDuration = tokenValue ? tokenValue.duration : setChainedValue(css.transitionDuration, parsedValue);
```

Please help me understand if this approach sounds good

vivek12345:
Have tried this with a sandbox here
https://codesandbox.io/s/react-stitches-forked-8rdye?file=/src/App.js

If the above approach looks good to us then I can update this PR with the changes.
For now the changes are in my forked repo
https://github.com/vivek12345/stitches/pull/5

vivek12345:
@hadihallak any advice here?

hadihallak:
@vivek12345 Thank you so much for all of the hard work on this PR and apologies for the late response  🙏  my notifications are all over the place so missed this one.

Yeah this seems to be working very well 👍  but I'm also wondering if we can at the same time support the longhand tokens object that @imjoshellis mentioned 🤔 


@jonathantneal what do you think? how do you think we should handle this?

jonathantneal:
In general, I like this, because I think eventually most properties will need to be parsed. I like that it’s creating a structure I suspect we could also apply to other properties. Is that right?

This is good stuff. 👍

jonathantneal:
One thing I hadn’t fully considered was that this current PR changes what the author has written, while the comment @vivek12345 represents something that preserves original intentions. Are you open to updating the PR so your latest suggestion may be reviewed more thoroughly?

vivek12345:
@jonathantneal yes you are absolutely right on this. I will update the pr today. The reason I didn't update it before was to get some feedback around my approach.

Thank you for taking the time to review the approach and I will update the pr today.

imjoshellis:
Just wanted to pop in again and give a more clear description of the API I had imagined because I'm not sure if I was very articulate after re-reading my original comment. 

With a `tokens` that looks like this:

```
tokens: {
  transitions: {
     slow: '1000ms',
     customEasing: 'cubic-bezier(.17,.67,.83,.67)'
  }
}
```

In the React styled composer, I could then do:

```
// ...
transition: "colors $slow $customEasing"
```

Which is similar to what I would naturally do as a workaround:

```
import {slow, customEasing} from '../cssVariables'

// ...
transition: `colors ${slow} ${customEasing}`
```

peduarte:
Hey, this is now working in the new beta canaries. Thanks

PR #184, opened 2020-09-05
Add config breakpoints and improve tests

### LINKED ISSUE #75: Improve testing

### LINKED ISSUE #94: Config media query in variant throws
Repro: https://codesandbox.io/s/stitches-poc-config-media-query-in-variant-3tm3r?file=/src/App.tsx

Notice:
- Referencing a media query from the `config.mediaqueries` throws an error

Expectation:
- Ability to use config media queries in variants (https://github.com/modulz/stitches/issues/86)

hadihallak:
@peduarte This will fix #94 by making nested breakpoints override breakpoints with lower nesting as we've discussed before.
also, this improves testing for styled-components to include media queries styles (see `index.test.tsx.snap` for an example) and with this fixed, the structure should be in place to test styled-components so this in addition to the work that we did a few days ago around testing and building should fix #75

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -157,11 +157,6 @@ const resolveBreakpointAndSelectorAndInlineMedia = (nestingPath: string[], confi
       }
       // breakpoints handling:
       if (breakpointOrSelector in config.breakpoints || breakpointOrSelector === MAIN_BREAKPOINT_ID) {
-        if (acc.breakpoint !== MAIN_BREAKPOINT_ID) {
-          throw new Error(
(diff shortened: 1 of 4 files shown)


peduarte (approved):

peduarte:
Great stuff, will do some testing!

PR #197, opened 2020-09-09
Fix global style injection in SSR

### LINKED ISSUE #185: Global styles not working with SSR
# Bug report

## Describe the bug

When using SSR using the following command

    yarn create next-app --example with-stitches

Global styles don't seem to work

## To Reproduce

1. `yarn create next-app --example with-stitches`
2. Add global styles
3. See them not getting applied

## Expected behavior

The global styles should be applied as they get on a project not using SSR

## System information

- OS: macOS
- Browser: Chrome
- Version of Next.js: 9.5.3

peduarte:
Fixes #185

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -228,7 +228,9 @@ const createCssRule = (breakpoints: IBreakpoints, atom: IAtom, className: string
     cssRule = `${createSelector(className, atom.selector)}{${atom.cssHyphenProp}:${atom.value};}`;
   }
 
-  return atom.breakpoint !== MAIN_BREAKPOINT_ID ? breakpoints[atom.breakpoint](cssRule) : cssRule;
+  return atom.breakpoint === MAIN_BREAKPOINT_ID || atom.breakpoint === '__global__'
+    ? cssRule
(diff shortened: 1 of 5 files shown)


peduarte:
breakpoints are not working. looking into it

peduarte:
I've written some tests for the cases that are not working. 

I also realised that the style object can be used in different ways. It seems that right now, you can do:

```js
css.global({
	body: {
		backgroundColor: 'red',
		'@media (min-width: 666px)': {
			backgroundColor: 'black',
		},
	},
});
```

**or**

css.global({
	body: {
		backgroundColor: 'red',
	},
	'@media (min-width: 666px)': {
		body: {
			backgroundColor: 'black',
		}
	}
});
```

And the both kinda work. 🤔 

@hadihallak may be able to provide some context on this, whether its on purpose or not.

But in any case, Im not sure I like it. 

Opinions?

peduarte:
Just realised we have the same issue with `css.keyframes` 🤔 

https://github.com/modulz/stitches/issues/203

peduarte:
Closing. New one coming :)

PR #201, opened 2020-09-09
fix: Fix check against Deno

### LINKED ISSUE #199: Check against server fails on Deno
# Bug report

Stitches fails to run through Deno.

## Describe the bug

I am currently trying to use stitches through Deno (1.3.3). I found out that there's an assumption in SSR that doesn't hold in the environment and subsequently running stitches through Deno crashes.

## To Reproduce

Make sure to install Deno first (available through brew for example). Then, run the sample script below:

```ts
import css from "https://dev.jspm.io/@stitches/css@9.0.0-alpha.8";

const c = css.createCss({ prefix: "", tokens: {}, breakpoints: {}, utils: {} });
```

It will crash with `error: Uncaught TypeError: Cannot read property 'documentElement' of undefined`.

## Expected behavior

It should not crash.

## Proposed solution

The issue occurs because the code checks for the environment based on the existence of `window`. In Node environment it's `undefined` but in Deno it's defined as it maintains partial compatibility with the object.

To solve, there should be a more robust check against the environment that works reliably for both Node and Deno. Maybe a simple way would be to check against the `Deno` object as that's guaranteed to exist for Deno but it shouldn't exist in the browser.

## System information

- OS: macOS
- Deno 1.3.3.

bebraw:
Since I know `window` object exists on Deno and it has `Deno` object within it, I added a check based on that.

Of course the question is how to test this effectively. 😄 

Closes #199.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -343,12 +343,19 @@ const composeIntoMap = (map: Map<string, IAtom>, atoms: (IAtom | IComposedAtom)[
   }
 };
 
+declare global {
+  interface Window {
+    Deno: any;
+  }
+}
+
 export const createTokens = <T extends ITokensDefinition>(tokens: T) => {
   return tokens;
 };
 export const createCss = <T extends TConfig>(
   _config: T,
-  env: Window | null = typeof window === 'undefined' ? null : window
+  // Check against Deno env explicitly #199
+  env: Window | null = typeof window === 'undefined' || window?.Deno ? null : window
 ): TCss<T> => {
   // pre-checked config to avoid checking these all the time
   // tslint:disable-next-line


peduarte:
@bebraw Thanks for this <3

While we're still improving our CI workflow and PRs arent running tests. Can you confirm that you can do these succesfully:
- `yarn test`
- `yarn build`

Otherwise this PR looks good to me!

bebraw:
@peduarte Yeah, both complete.

If possible, I would consider adding Deno to the build matrix of your CI once you set it up. That way we can see how it goes with the compatibility. I'm not sure if that's trivial, though.

peduarte:
@bebraw sounds like a good idea, thanks for that

peduarte (approved):

PR #166, opened 2020-09-03
update props of a styled component for better DX

deebov:
Before and After explanation:
https://streamable.com/llq74k

PR #207, opened 2020-09-10
(suggestion) feat: withComponent API

cometkim:
Implemented `withComponent()` API 

- Emotion docs: https://emotion.sh/docs/styled#change-the-rendered-tag-using-withcomponent
- styled-comopnents docs: https://styled-components.com/docs/api#withcomponent

styled-components says it's going to be deprecated, but it's still worth it. It's easier to understand than as prop, and **TypeScript typing is more stable and fast.**

Uses can reuse styles while simultaneously **restricting bindings from being dynamically changed at runtime.**

It is a useful tool for defining components for the same styles and adds only 70 characters of code based on minified results. (even better result if it reuses code with the proxy below)

```tsx
const Icon = styled('svg', {
  width: '24px',
  height: '24px',
});

const FacebookIcon = Icon.withComponent(FacebookSvg);
const TwitterIcon = Icon.withComponent(TwitterSvg);

<FacebookIcon />
<TwitterIcon />
{/* vs */}
<Icon as={FacebookSvg} />
<Icon as={TwitterSvg} />

});
const ButtonLink = Button.withComponent(Link);

<ButtonLink />
{/* vs */}
<Button as={Link}/>
```

### DIFF
--- a/packages/react/src/index.tsx
+++ b/packages/react/src/index.tsx
@@ -162,9 +162,15 @@ export const createStyled = <Config extends TConfig>(
 
   const configBreakpoints = config.breakpoints || {};
 
+  type EvaluatedVariantMap = Map<string, Map<string, { [key: string]: string }>>;
+  type EvaluatedCompoundVariants = Map<any, { [key: string]: string }>;
   const styledInstance = (
     baseAndVariantStyles: any = (cssComposer: any) => cssComposer.compose(),
-    Component: React.ComponentType<any> = Box
+    Component: React.ComponentType<any> = Box,
+    prevState?: {
+      evaluatedVariantMap: EvaluatedVariantMap;
+      evaluatedCompoundVariants: EvaluatedCompoundVariants;
+    }
   ) => {
     let numberOfCompoundVariants = 0;
     const as = currentAs;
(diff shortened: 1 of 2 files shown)


peduarte:
@cometkim thanks so much for taking the time to suggest this

One of Stitches goals is to keep the API surface as thin as possible. 

We spent along time discussing whether between the `as` prop and `withComponent` already, and we've decided to stick with the `as` prop.

Main reason being is that the majority of people preferred to see the rendering element inline with the context. 

In any case, we already have a few ways of achieve this:
```js
const Button = styled('button', {})

// 1.
const ButtonLink = styled(Button, {})
ButtonLink.defaultProps = { as: 'a' }

// 2.
const ButtonLink = props => <Button as="a" {...props} />

// 3. 
<Button as="a" />

```

Until we reach stable v1 I'd like to refrain from introducing new APIs, unless totally necessary.

> TypeScript typing is more stable and fast.

I don't know much about this, but this could be a legit reason to introduce it. If we can run some benchmarks we can add this under consideration. 

But, we can't drop the `as` prop, that is the best developer experience for majority of users.

Cheers!

cometkim:
As props in many cases fail inference and the remaining props are extremely slow to compute.

for example in stitches-site:
image

If we can replace the tag without using the as prop, We can avoid of inferring the rest of the prop from as tag because the inference ends at the `as?: never`.

peduarte:
@cometkim we've these [typing improvements](https://github.com/modulz/stitches/pull/166), but not released them yet. 

This codesandbox should be using that already: https://codesandbox.io/s/stitches-ci-nextjs-ufdek?file=/pages/index.tsx

cometkim:
```
// 1.
const ButtonLink = styled(Button, {})
ButtonLink.defaultProps = { as: 'a' }
```
This doesn't match the use case.

```
const TwitterIcon = styled(Icon, {})
TwitterIcon.defaultProps = { as: TwitterSvg };

const FacebookIcon = styled(Icon, {})
TwitterIcon.defaultProps = { as: FacebookSvg };
```
And corrected example is a little verbose.

```
// 2.
const ButtonLink = props => <Button as="a" {...props} />

// 3. 
<Button as="a" />
```
And this example is still using the as prop I wanna avoid using.

There is an assumption that withComponent will not change the style, so we can skip expensive calculations when copying via some internal flags.

And unlike the as prop, it's already efficient because it's mostly done outside of render.

cometkim:
@peduarte Hmm, I am wondering we already have a way to avoid cloning of internal instance on rerendering when using the as prop static.

jjenzz:
`defaultProps` is being deprecated (https://github.com/facebook/react/pull/16210) so, seems we should try to come up with a first-class approach to this whatever it may be.

martonlanga:
Not sure whether making TS work with these is possible, but here are some ideas for a `defaultProps` replacement:

### 1.
```ts
  variants: {
    color: {
      violet: {
        default: true,
        backgroundColor: 'blueviolet',
      },
      gray: {
        backgroundColor: 'gainsboro',
      },
    },
  },
});


render(<Button>Button</Button>);
```

### 2
```ts
  defaultProps: {
    color: 'violet'
  },
  variants: {
    color: {
      violet: {
        backgroundColor: 'blueviolet',
      },
      gray: {
        backgroundColor: 'gainsboro',
      },
    },
  },
});


render(<Button>Button</Button>);
```

### 3
```ts
  variants: {
    color: {
      violet: {
        backgroundColor: 'blueviolet',
      },
      gray: {
        backgroundColor: 'gainsboro',
      },
    },
  },
}, { color: 'violet' }); // third arg for defaultProps


render(<Button>Button</Button>);
```

As a user, I'd prefer the last one. Would be also ideal for composition:
```ts
const ButtonLink = styled(Button, {}, { color: 'gainsboro', as: Link })
```

Also, I think the following is nicer than `withComponents`:
```ts
const CardLink = (props) => <Card as="a" {...props} />
```

cometkim:
This is well inferred as correct type without any annotations:
```tsx
const CardLink = Card.withComponent(Link);
```

This is way more verbose and also expensive to infer:
```tsx
const CardLink = (props: React.ComponentProps<typeof Card>) => <Card as={Link} {...props} />
```

and inference could fail when the card itself is already polymorphic.

cometkim:
As mentioned above, I made some evaluations skipped when cloning with withComponent. This is safe as it only changes only the element and reuses the style.

peduarte:
@cometkim thanks for all this. much appreciated. I've opened up a discussion which is somewhat related to this: https://github.com/modulz/stitches/discussions/242

peduarte:
@cometkim we're doing a bunch of refactoring and will consider this. But this PR won't work with our new codebase. Thanks so much <3

PR #227, opened 2020-09-18
Add interface for numeric tokens

### LINKED ISSUE #226: Support numeric values for tokens
Can we have numeric values allowed for tokens like `fontWeights, `zIndices`, `linHeights`, etc?
Example:  https://codesandbox.io/s/stitches-ts-issue-forked-voh43?file=/src/stitches.config.ts

hauptrolle:
Fixes #226 by adding a new type for numeric tokens

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -362,20 +362,25 @@ export interface ITokenDefinition {
   [key: number]: string;
 }
 
+export interface INumericTokenDefinition {
+  [key: string]: number;
+  [key: number]: number;
+}
+
 export interface ITokensDefinition {
   colors?: ITokenDefinition;
   space?: ITokenDefinition;
   fontSizes?: ITokenDefinition;
   fonts?: ITokenDefinition;
-  fontWeights?: ITokenDefinition;
-  lineHeights?: ITokenDefinition;
+  fontWeights?: INumericTokenDefinition;
+  lineHeights?: ITokenDefinition | INumericTokenDefinition;
   letterSpacings?: ITokenDefinition;
   sizes?: ITokenDefinition;
   borderWidths?: ITokenDefinition;
   borderStyles?: ITokenDefinition;
   radii?: ITokenDefinition;
   shadows?: ITokenDefinition;
-  zIndices?: ITokenDefinition;
+  zIndices?: INumericTokenDefinition;
   transitions?: ITokenDefinition;
 }
 export interface IUtils {
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -367,12 +367,34 @@ export interface INumericTokenDefinition {
   [key: number]: number;
 }
 
+type TokenKey = string | number;
+
+export type IFontWeightTokenDefinition = {
+  [key in TokenKey]:
+    | 'normal'
+    | 'bold'
+    | 'bolder'
+    | 'lighter'
+    | 100
+    | 200
+    | 300
+    | 400
+    | 500
+    | 600
+    | 700
+    | 800
+    | 900
+    | 'inherit'
+    | 'initial'
+    | 'unset';
+};
+
 export interface ITokensDefinition {
   colors?: ITokenDefinition;
   space?: ITokenDefinition;
   fontSizes?: ITokenDefinition;
   fonts?: ITokenDefinition;
-  fontWeights?: ITokenDefinition | INumericTokenDefinition;
+  fontWeights?: IFontWeightTokenDefinition | ITokenDefinition | INumericTokenDefinition;
   lineHeights?: ITokenDefinition | INumericTokenDefinition;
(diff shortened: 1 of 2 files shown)


packages/core/src/types.ts
```diff
@@ -362,20 +362,25 @@ export interface ITokenDefinition {
...
+}
+
 export interface ITokensDefinition {
   colors?: ITokenDefinition;
   space?: ITokenDefinition;
   fontSizes?: ITokenDefinition;
   fonts?: ITokenDefinition;
-  fontWeights?: ITokenDefinition;
-  lineHeights?: ITokenDefinition;
+  fontWeights?: INumericTokenDefinition;
```
afzalsayed96, line 375: It would be nice to have `ITokenDefinition | INumericTokenDefinition` here

afzalsayed96: I could not only provide `600` as a value but also `"600"` or `"bold"` as a string

packages/core/src/types.ts
```diff
@@ -362,20 +362,25 @@ export interface ITokenDefinition {
...
+  lineHeights?: ITokenDefinition | INumericTokenDefinition;
   letterSpacings?: ITokenDefinition;
   sizes?: ITokenDefinition;
   borderWidths?: ITokenDefinition;
   borderStyles?: ITokenDefinition;
   radii?: ITokenDefinition;
   shadows?: ITokenDefinition;
-  zIndices?: ITokenDefinition;
+  zIndices?: INumericTokenDefinition;
```
afzalsayed96, line 383: Same here. Changing it to `number` only would make it a breaking change.
Let's allow for both numeric as well as string types

packages/core/src/types.ts
peduarte: Just an idea: I think eventually it would be nice to have scale types. so for font weights it would suggest the keywords (`bold`) and numbers (`100-900`). For colors, it could suggest the color keywords (`tomato`, `rebeccapurple`).

afzalsayed96:
@peduarte If it's not too much to ask can we have a patch released tonight?
Can't wait to get my hands on this 😄

peduarte:
@afzalsayed96 haha sorry, that wont be possible, unfortunately. Theres has been a few things going in and I need to test them before releasing. 

But if you really want this now you can set your package.json to:

```
"@stitches/react": "https://pkg.csb.dev/modulz/stitches/commit/449f5533/@stitches/react",
```

peduarte (commented):
Good stuff! It's looking good. 

Regarding the types, I think that'll do for now, but ideally I think we should rely on the `FontWeightAbsolute` from `css-types`.

Can you write some a test to validate these changes?

Thanks!

hadihallak (commented):
@hauptrolle Thanks for submitting this PR. Looks great!

@peduarte one note, I think we should disallow the token **keys** from being number as using them number value instead of the string representation in the style object will not yield the token as they need to be strings. what do you think?

packages/core/src/types.ts
jonathantneal: For `font-weight`, the numeric value can accept any numbers between 1 and 1000. My current understanding is that TypeScript will not allow union types this big.

- [Reference: microsoft/TypeScript - Suggestion: Range as Number type](https://github.com/microsoft/TypeScript/issues/15480)

peduarte: @jonathantneal Yeah, 1 to 1000 is a bit much :D From what Ive seen, 100-900 is the most common

peduarte:
@hadihallak yeah, agreed.

hauptrolle:
Hey guys :wave:

to sum it up, we need to

- add some unit tests for these changes (I could use some help here)
- update the `FontWeightAbsolute` from css types and use this instead of defining a new one
- Don't allow token keys to be a number (Would't this be a breaking change?)

hadihallak:
> add some unit tests for these changes (I could use some help here)

working on enhancing testing types later but for now, just a test case in the test file should work fine and will cause the tests to fail if there were type errors.
if you want to assert an error you could use this line comment before the line that you expect to error:
```
  // @ts-expect-error
  as: 'potato',
```
> Don't allow token keys to be a number (Would't this be a breaking change?)

Tokens keys already need to be strings in order for them to work so this change would be fixing a TS bug instead of a breaking change

hadihallak:
Hey, @hauptrolle 👋 
let me know if you need any help with this one. should be ready for merge once the change requests are addressed ❤️

peduarte:
@hadihallak I had a chat with @hadihallak is he's been quite busy lately, so I suggest you take over if that's ok. Thanks! <3

afzalsayed96:
Yo! What happened to this PR? 😄

hadihallak:
Hey @afzalsayed96 👋 
This is on the radar and will be handled as part of #294

peduarte:
Closing in favour of #294

PR #230, opened 2020-09-20
refactor: removed unused tsconfig.build.json

andykenward:
Removed the the tsconfig.build.json configuration files as it appears they arent being used in the project.


peduarte (approved):
That's awesome. I really love these types of refactoring. Thanks!

packages/react/tsconfig.json
```diff
@@ -1,4 +1,7 @@
 {
   "extends": "../../tsconfig.json",
-  "include": ["src/**/*", "tests/**/*"]
+  "compilerOptions": {
+    "jsx": "react",
+  },
+  "include": ["src/**/*"]
```
peduarte, line 6: just one question: why did you remove tests from here?

andykenward: As the project is using Jest along with ts-jest to transform the TypeScript files which gives Jest allows typechecking to work.
Without ts-jest then you wouldn't get typechecking when running the tests as Jest uses Babel for TypeScript transforming.

But we should probably add `"test/**/*"` back in the `include` array for both `ts.config.json` in the packages folders incase you want to use TypeScript directly via the cli `yarn typecheck`.

peduarte: @andykenward gotcha. so I'll wait for you to add it back before merging. Cheers!

peduarte (approved):

packages/react/tsconfig.json
andykenward: @peduarte Done.

But if the project ever changes the build setup it may need removing.

hadihallak (approved):

hadihallak:
Thanks @andykenward 
loving the refactors

PR #231, opened 2020-09-20
refactor: jest configuration

andykenward:
- allows Jest to be run from the root of the project.  As Jest supports [`projects`](https://jestjs.io/docs/en/configuration#projects-arraystring--projectconfig) we can use that to run Jest from root of the project folder. _"This is great for monorepos"_. 
  - Also allows the usage of [VSCode Extension Jest](https://marketplace.visualstudio.com/items?itemName=Orta.vscode-jest) to work without custom configuration.

- removed what appears to be a legacy node version logic in [`jest.config.base.js`](https://github.com/modulz/stitches/compare/canary...andykenward:jest-config?expand=1#diff-66d3521f5e9f81934077fdd6653d0901L1-L18). We can use the projects tsconfig.json. Perhaps we need to define the package.json [`engines`](https://docs.npmjs.com/files/package.json#engines) for what version of node & yarn this project supports.
- generates coverage only on root `yarn test` and not in each package folder
- As yarn workspaces is used in the project there should be no need to reference Jest from the node_modules directory for the `test:watch` cli. ~~`node ../../node_modules/jest/bin/jest.js`~~ becomes `jest`.
- Removed some config and arguments that are already defined as the defaults for jest. e.g `--env=jsdom`
- Added root cli `yarn test:watch` so you can watch all test files in project.
- fix jest displayName to match package name.

### DIFF
--- a/jest.config.base.js
+++ b/jest.config.base.js
@@ -1,36 +1,8 @@
-const semver = require("semver");
-
-function getSupportedTypescriptTarget() {
-  const nodeVersion = process.versions.node;
-
-  if (semver.gt(nodeVersion, "10.0.0")) {
-    return "es2018";
-  } else if (semver.gt(nodeVersion, "7.6.0")) {
-    return "es2017";
-  } else if (semver.gt(nodeVersion, "7.0.0")) {
-    return "es2016";
-  } else if (semver.gt(nodeVersion, "6.0.0")) {
-    return "es2015";
-  } else {
-    return "es5";
-  }
-}
-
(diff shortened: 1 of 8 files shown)


peduarte:
Looks great to me. The simpler the better 😄 

@hadihallak and @christianalfoni do you guys mind taking a look at this one? Just wanna make sure that we're not removing something that was there for some edge case reason.

hadihallak (approved):
Working great on my end. Thanks @andykenward

PR #232, opened 2020-09-20
chore: remove dts-bundle npm package

andykenward:
Doesn't appear that the npm package [`dts-bundle`](https://github.com/TypeStrong/dts-bundle) is used in the project. 

Can we remove this?


peduarte:
Thanks!

@hadihallak can you take a look at this? Im not sure this was part of the work you did around the build process, etc. But yeah, looks like it isn't being used.

hadihallak (approved):
Nice one! thanks for submitting it ❤️

PR #235, opened 2020-09-22
Return string based className from styled components in the react package

### LINKED ISSUE #191: Stitches with Ionic components
# Bug report

## Describe the bug
I've been playing around with Stitches for various use cases, and so far so good. However I reached a point that was a "tipping point" for me to use it over styled-components permanently.

I cannot use it with Ionic Components example below. Maybe I am misunderstanding the docs, but it should accept React components.

```

const ToolBar = styled(IonToolbar, {
    display: "flex",
    alignItems: "center",
    justifyContent: "space-between",
    svg: {
        position: "relative",
        zIndex: 2,
        color: "#000"
    }
})
```

## To Reproduce
Setup a standard Ionic with React project -> https://ionicframework.com/docs/intro/cli
Setup Stitches
Try and create a stitched component similar to my code above.

## Expected behavior
Should wrap the Ionic component so I can style them with either direct styles or css properties.

## Screenshots
image

## System information

- OS: macOS
- Browser: Chrome
- Version of Node.js: 12.8.0

### LINKED ISSUE #229: Stitches do not work with React libraries that expect className to be a string
# Bug report

## Describe the bug

Stitches do not work correctly with some libraries that work with `className` prop because they expecting `className` prop to be a `string` but in Stitches, it's `object` with atoms and `toString` method.

And some libraries and even technologies (stencil.js - https://github.com/modulz/stitches/issues/191) work with className in the following way:

```javascript
className.split(' ')
```
- [React Font Awesome codebase](https://github.com/FortAwesome/react-fontawesome/blob/master/src/components/FontAwesomeIcon.js#L16)
- [StencilJS codebase](https://github.com/ionic-team/ionic-framework/blob/master/packages/react/src/components/utils/attachProps.ts#L40)

but this gives us an error because `split()` is not a method on an object that Stitches passes to the `className` prop. One solution to this problem would be to use `toString()` method and then use `split()` like this:

```javascript
className.toString().split(' ')
```
but this solution requires to change a lot of libraries and technologies out there. So it's not ideal.

## To Reproduce

[CodeSandbox with specific error](https://codesandbox.io/s/stitches-font-awesome-2nkx7?file=/src/App.tsx)

## Expected behaviour

Stitches will correctly work with libraries like [React Font Awesome](https://github.com/FortAwesome/react-fontawesome).

## System information

- OS: Windows
- Browser: any

hadihallak:
fixes #191 & fixes #229

### DIFF
--- a/packages/react/src/index.tsx
+++ b/packages/react/src/index.tsx
@@ -257,7 +257,7 @@ export const createStyled = <Config extends TConfig>(
         ...propsWithoutVariantsAndCssProp,
         as: props.as || as,
         ref,
-        className: css(stitchesComponentId, ...compositions, props.className),
+        className: css(stitchesComponentId, ...compositions, props.className).toString(),
       });
     });
     StitchesComponent.displayName =
--- a/packages/react/tests/index.test.tsx
+++ b/packages/react/tests/index.test.tsx
@@ -62,7 +62,7 @@ expect.addSnapshotSerializer({
   },
 
   test(val) {
-    return val && val.props && val.props.className[_ATOM] && val.children && val.type;
+    return val && val.props && val.props.className && val.children && val.type;
   },
 });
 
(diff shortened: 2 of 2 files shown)


olidotjpeg:
Super cool to get this fixed! Excited for it :D

peduarte (approved):

peduarte:
Nice one. Will include this in the next canary release.

peduarte:
@christianalfoni mind having a quick look? I'm curious if there are any downsides to this. Cheers

PR #238, opened 2020-09-22
Fix numeric 0 value variant not working

hadihallak:


peduarte:
Getting some TS errors during usage: https://codesandbox.io/s/stitches-ci-nextjs-forked-nuizt?file=/pages/index.tsx

peduarte (changes requested):
Getting some TS errors during usage: https://codesandbox.io/s/stitches-ci-nextjs-forked-nuizt?file=/pages/index.tsx


CleanShot 2020-09-22 at 20 40 58@2x

hadihallak:
this PR doesn't alter the types in any way so i'm not sure if this is related O.o 
is this on codesandbox or locally?

hadihallak:
@peduarte just checked and the error exists in older PRs so this typing issue was probably introduced somewhere else

peduarte:
Yeah this always existed. But I see it as kind of related to this one. How big is it to fix?

hadihallak:
Just pushed a fix for it so now it additionally accepts string values when numbers are used as variants

sadly, there doesn't seem to be a way to cast a number to its string representation so this will  only cast it to string

so a numeric variant `1` will be accept `1` | string instead of `1` |`"1"`

peduarte (approved):
LGTM, but is it worth writing a test for this?

peduarte (approved):

PR #245, opened 2020-09-25
Feature/comma separated rules & fixed ampersand support

### LINKED ISSUE #233: Styles aren't scoped in multiple pseudo-classes
# Bug report

## Describe the bug

When multiple pseudo-classes specified in `variants`, only first one is scoped to the component, second one becomes global.

```tsx
  padding: "14px 20px",
  variants: {
    variant: {
      blue: {
        ":focus, :active": {
          color: "blue"
        }
      }
    }
  }
});
```

This will generate something similar to:

```
._initial_c_bFQQno._initial_c_bFQQno._initial_c_bFQQno:focus, :active {
    color: blue;
}
```
(when `:focus` or `:active`)

So, the last `:active` gets applied globally.

## To Reproduce

Here's a [sandbox](https://codesandbox.io/s/stitchesreact-leaking-styles-ieseh?file=/src/App.tsx)

## Expected behavior

All pseudo-classes scoped to the variant:

```
.generated:focus, .generated:active {
    color: blue;
}
```

hadihallak:
**Two new things:**
- ampersand should now work like in other css in js library when nested - fixes #233
- comma-separated rules will work and be scoped correctly

Please it a good old street test and let me know if you find any edge case that needs handling

The tests diff is a bit large as the selectors got reworked which caused the hashing function to return a different hash but other than that I've added new tests cases to test the new functionality

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -39,7 +39,7 @@ const createSSRCssRuleClass = (s: string) => {
 
 const createSelector = (className: string, selector: string) => {
   const cssRuleClassName = className ? `.${className}` : '';
-  if (selector && selector.includes('&')) return selector.replace(/&/gi, cssRuleClassName);
+  if (selector && selector.includes('%')) return selector.replace(/\%/gi, cssRuleClassName);
   if (selector) {
     return `${cssRuleClassName}${selector}`;
   }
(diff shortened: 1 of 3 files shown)


peduarte (approved):
LGTM, will need some proper testing though. I'll get this in Canary later today

PR #251, opened 2020-09-30
Add `as` defaultProps types and tests

### LINKED ISSUE #223: Missing element types when using `as` in `defaultProps`
# Bug report

## Describe the bug

The element types are missing when `as` is used in `defaultProps`. It works when used inline

## To Reproduce

https://codesandbox.io/s/stitches-ts-issue-forked-op7gi?file=/src/App.tsx

## Expected behavior

The types should reflect the element props passed to the `as` prop, both inline and in `defaultProps`

hadihallak:
Fixes #223

Note: there is diff noise in the tests related to the changes we did in the core package for the comma-separated rules and ampersand in #245

### DIFF
--- a/packages/react/src/index.tsx
+++ b/packages/react/src/index.tsx
@@ -91,7 +91,7 @@ export interface IStyledComponent<ComponentOrTag extends React.ElementType, Vari
   /**
    * Default props typing:
    */
-  defaultProps?: VariantASProps<Config, Variants> & { [k: string]: any };
+  defaultProps?: VariantASProps<Config, Variants> & { as?: React.ElementType } & { [k: string]: any };
   /**
    * DisplayName typing:
    */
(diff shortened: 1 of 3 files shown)


hadihallak:
@peduarte just tested against your use case in the original issue and realized you wanted the usage in jsx to reflect the changes you made in the defaultProps which isn't possible in typescript with the syntax you wrote.

if you need something similar then we'd need to make the default props a function call which then we could use to return a new component with the default props and their typing applied to them. (I think you showed me something like this previously in one of [your discussions](https://github.com/modulz/stitches/discussions/242))

**so for this to work we need:**
```jsx
const Text = styled("p", {});

const TextLink = styled(Text, {}).withProps({ as :'a'});
```

**or:** 

```jsx
const Text = styled("p", {});

const BaseTextLink = styled(Text, {});
const TextLink = BaseTextLink.withProps({as: 'a'})
```

in both of the examples above, we could use the withProps **function** to return a new component with its appropriate typing

**other alternatives:**

```jsx
import {applyDefaultProps} from '@stitches/react'

const Text = styled("p", {});

const BaseTextLink = styled(Text, {});
const TextLink = applyDefaultProps(BaseTextLink, {as: 'a'})
```
**or:**
```jsx
import {applyDefaultProps} from '@stitches/react'

const Text = styled("p", {});

const TextLink = styled(Text, {...styles}, {...defaultProps});
```

peduarte (approved):

peduarte:
LGTM!

PR #254, opened 2020-10-01
Update `flex` parser

peduarte:
This PR fixes the parsing of `flex` property

### DIFF
--- a/packages/core/src/shorthand-parser/index.ts
+++ b/packages/core/src/shorthand-parser/index.ts
@@ -381,7 +381,15 @@ export const flex = (tokens: any, value: any) => {
   if (parsedValue.length === 2) {
     return {
       flexGrow: parsedValue[0],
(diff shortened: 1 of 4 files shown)


jonathantneal (commented):
- [x] Initial code review 👍

jonathantneal (approved):
- [x] Code review 👍
- [x] Checkout review 👍

🚢 Love it! ♥❤️﻿

hadihallak (approved):
LGTM ✅ 

@jonathantneal 
I think we should start to experiment with moving the shorthand parsing outside the library as a plugin of some sort because all throughout our shorthand parsing we have cases like this where we're not following the CSS spec 1:1 and we're not generating full longhands that include setting defaults and since doing so we are likely to increase the bundle size then it would be very good if we could separate this into its own plugin or package.
curious to know your thoughts about this

jonathantneal:
@hadihallak strong agree, next canary? 😄 let’s discuss next.

PR #258, opened 2020-10-02
Fix and test class name composition

### LINKED ISSUE #256: Specificity changed after upgrading to v0.0.3-canary.2
# Bug report

## Describe the bug

After upgrading to latest canary version specificity seems to have changed.

## To Reproduce

1. Check this sandbox: https://codesandbox.io/s/heuristic-zhukovsky-pec21?file=/src/App.tsx

## Expected behavior

Link color should have been $purple600 but it stays $gray600

peduarte:
This PR fixes class name composition. I've added a couple of extra tests to catch this in the future.

Closes #256

### DIFF
--- a/packages/react/tests/__snapshots__/index.test.tsx.snap
+++ b/packages/react/tests/__snapshots__/index.test.tsx.snap
@@ -136,6 +136,21 @@ exports[`styled Renders basic component 1`] = `
 }
 `;
 
+exports[`styled Renders components from other libraries 1`] = `
+{
+  "orderedAppliedStyles": [
+    "._vfarC {color: red;}",
+    "._vfarC {color: red;}"
+  ],
+  "props": {
+    "className": "scid-gmqXFB _vfarC "
+  },
+  "children": [
+    "hello world"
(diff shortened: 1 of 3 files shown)


packages/react/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -136,6 +136,36 @@ exports[`styled Renders basic component 1`] = `
 }
 `;
 
+exports[`styled Renders components from other libraries 1`] = `
+{
+  "orderedAppliedStyles": [
+    "._vfarC {color: red;}",
+    "._vfarC {color: red;}"
```
peduarte, line 143: This looks a bit off. It's rendering it twice. Is this normal? @hadihallak

hadihallak: good catch 💯 
I'm hoping this is a bug with the jest serializer but if not then there might be an issue with the atom caching functionality

hadihallak (approved):

PR #260, opened 2020-10-05
Enhance/compund variants responsive

hadihallak:


tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "ES6",
     "module": "commonjs",
-    "lib": ["dom", "es2017"],
+    "lib": ["dom", "es2019"],
```
peduarte, line 5: these changes will impact how the library is built, right? is that on purpose?

packages/react/tests/index.test.tsx
```diff
@@ -337,4 +337,58 @@ describe('styled', () => {
...
+      }
+    );
+
+    expect(
+      renderer
+        .create(
+          <Button color={{ initial: 'red', breakpointOne: 'yellow' }} size={{ breakpointTwo: 'big' }}>
+            Responsive compound variants
+          </Button>
+        )
```
peduarte, line 389: Nice! this works as I would expect

This test is a bit complex though, since it relies on an implicit breakpoint match.

Do you think its worth writing a few different tests for the compound API, for the sake of clarity?

- the most basic usage
- a usage where the breakpoint match is explicit
- a usage where the breakpoint match is implicit (this one)

hadihallak: Updated the tests 👍

tsconfig.json
hadihallak: yeah to get the latest features like flatMap so if anything, this will make the build smaller but will double-check that this will still work on all modern browsers

hadihallak:
@peduarte Mind giving this another test?
fixed a few issues with the generation and added better tests

peduarte (commented):
Tests are looking good! 🔥 

There's a `.only` though :D

packages/react/tests/index.ssr.test.tsx
```diff
@@ -0,0 +1,267 @@
...
+      return '';
+    });
+    expect(styles.length).toBe(6);
+    expect(styles[5]).toMatchInlineSnapshot(`
+      "/* STITCHES:_COMPOUND_VARIANTS */
+      ./*X*/__COMPOUND_VARIANTS_bc_fhWNnZ/*X*/ {background-color:compoundColor;}"
+    `);
+  });
+
+  test.only('it matches non-implicit initial values', () => {
```
peduarte, line 219: `.only`

peduarte:
@jjenzz mind having a quick look? You can test it by forking [this codesandbox](https://codesandbox.io/s/stitches-ci-cra-forked-c2bb1?file=/src/stitches.config.ts)

jjenzz:
I expected to be able to do this, but it didn't work:

```jsx
<Button color="gray" appearance={{ initial: 'outline', bp1: undefined }}>
  Button
</Button>
```

I was expecting to see the compound variant applied intitially, but then just the "grey" button from `bp1` onwards. I tried to pass `null` but types moaned but I'm wondering if we should allow that when people want to explicitly unset something like this (or maybe we could even use [`unset`](https://developer.mozilla.org/en-US/docs/Web/CSS/unset)).

peduarte:
@hadihallak yeah @jjenzz has a point. this is her sandbox https://codesandbox.io/s/stitches-ci-cra-forked-06ekz?file=/src/App.tsx:1507-1514

hadihallak:
hmm.. not sure how this case would even be handled in normal CSS. 
the initial breakpoint is matching so the CSS rules have every reason to be present on larger breakpoints not unless we reset them (which would need in even more atoms that would do the job of reversing the styles on larger breakpoints)

I'm not sure if this is possible without generating even more atoms to do the reset but before I jump to conclusions, could you please provide me with an example of how you would handle this with pure CSS? like on what bases do we reset the styles when `bp1: undefined`?

jonathantneal:
Given `appearance = { initial: 'outline', bp1: undefined }`; is the interpretation being considered that `appearance.initial` would represent a “_catch all_” breakpoint key that matches any media condition not already matched by the other breakpoint keys on `appearance`?

hadihallak:
it would just generate rules with no @media rules at all so they will be present at all times

when  a breakpoint like bp1 or bp2 is involved then they would be present as long as their media queries match

jjenzz:
> not sure how this case would even be handled in normal CSS.

Haha me either. There's definitely a disparity between expectation and possibility here 😅 

This was my thought process:

- no breakpoint: There is a compound specified as `color: 'gray'` and `appearance: 'outline'` which matches a compound, so we apply it
- bp1: There is a compound specified as `color: 'gray'` and `appearance: null` which *does not* match a compound, so we apply `gray` styles only

I realise the result I'm seeing is because it is falling back to `initial: 'outline'` which I totally expect **if I'm not compounding**, but I wouldn't pass `bp1: null` if I wasn't compounding either.

For some reason, adding compound variants to the mix seemed to create the expectation that I should be able to cancel out a compound match at a different breakpoint. But I guess the consumer could just create a `max-width` breakpoint in this case and use that instead of `initial` 🤔 

I'm not sure now haha... this all hurts the brain!

hadihallak:
@peduarte I'm thinking we should probably delay this until after v1 as I'm starting to think that we will need some different compound variants API and I'm not sure that this new implementation checks all the boxes, in fact, it introduces more complexities now that variants can span multiple breakpoints at the same time

For now, I'd vote that we document compound variants better to showcase how they work in their current shape(without compound variants spanning multiple breakpoints) until we can come up with an API that checks all the boxes

Maybe a discussion around what compound variants are, the API, and how they deal with responsive breakpoints in all cases

peduarte:
@hadihallak Ok, I'll schedule a call about Compound API with yourself and @jonathantneal 

We need this to be solid before v1 though 🚀

peduarte:
closing this in favour of OOCSS by @jonathantneal we'll need to talk about this

PR #264, opened 2020-10-07
Add way to extract props & variants types

### LINKED ISSUE #248: React.ComponentProps not working as expected
# Bug report

## Describe the bug

When wanting to get the type of the props a stitches component receives, for example to be able to compose other components on top of it, `React.ComponentProps<typeof StitchesComponent>` is not working as expected — it's signaling weird errors and not providing correct typings. More context [in this discussion](https://github.com/modulz/stitches/discussions/213).

## To Reproduce

1. Go to [this sandbox](https://codesandbox.io/s/stitches-ts-issue-8stcu?file=/src/App.tsx).
2. Scroll down to the implementetion of `<LoadingButton />`.
3. See that `<LoadingButton />` has type errors. (property `as` is required).
4. Also, see that the props are not button-specific (see screenshot below).

## Expected behavior

I expected LoadingButton to be error free with the props provided, and for prop suggestions to be button-specific.

## Screenshots

CleanShot 2020-09-26 at 10 54 37@2x

hadihallak:
Adds ExtractVariantProps generic util that allows extracting the variant props out of stitches components
Allows React.ComponentProps to work with stitches components - Fixes #248

### DIFF
--- a/packages/react/src/index.tsx
+++ b/packages/react/src/index.tsx
@@ -74,6 +74,7 @@ export interface IStyledComponent<ComponentOrTag extends React.ElementType, Vari
     // e.g. some HTML elements have `size` attribute. And when you combine
     // both types (native and variant props) the common props become
     // unusable (in typing-wise)
+
     props: MergeElementProps<As, VariantASProps<Config, Variants>> & {
       as?: As;
       css?: TCssWithBreakpoints<Config>;
@@ -118,6 +119,20 @@ export type TProxyStyledElements<Config extends TConfig> = {
     a: BaseAndVariantStyles | TComponentStylesObject<Config>
   ) => IStyledComponent<key, TExtractVariants<BaseAndVariantStyles>, Config>;
 };
+
+export type TExtractVariantProps<C> = C extends IStyledComponent<infer T, infer V, infer G>
+  ? VariantASProps<G, V>
+  : never;
+
+export type TStyledComponentProps<C> = C extends IStyledComponent<infer T, infer V, infer G>
+  ? MergeElementProps<T, VariantASProps<G, V>> & {
+      as?: T;
+      css?: TCssWithBreakpoints<G>;
+      className?: string;
+      children?: any;
+    }
+  : never;
+
 /**
  * Styled Components creator Type.
  * ie: styled.div(styles) | styled('div', {styles})
--- a/packages/react/src/index.tsx
+++ b/packages/react/src/index.tsx
@@ -39,6 +39,13 @@ export type CastStringToBoolean<Val> = Val extends 'true' | 'false' ? boolean | 
 export type CastNumberToString<Val> = Val extends number ? string & {} : never;
 
 /**
+ * Extract variant props from Stitches components
+ */
+export type ExtractVariantProps<C> = C extends IStyledComponent<infer T, infer V, infer G>
+  ? VariantASProps<G, V>
+  : never;
+
+/**
  * Takes a variants object and converts it to the correct type information for usage in props
  */
 export type VariantASProps<Config extends TConfig, VariantsObj> = {
--- a/packages/react/src/index.tsx
+++ b/packages/react/src/index.tsx
@@ -41,8 +41,19 @@ export type CastNumberToString<Val> = Val extends number ? string & {} : never;
 /**
  * Extract variant props from Stitches components
  */
-export type ExtractVariantProps<C> = C extends IStyledComponent<infer T, infer V, infer G>
-  ? VariantASProps<G, V>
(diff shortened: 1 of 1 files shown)


peduarte:
@hadihallak cool, nice workaround. A couple of questions:
- If you manage to solve this with `React.ComponentProps` would you still keep these? Asking so I can manage breaking change expectation
- Is it worth exposing a `StitchesComponent` type that does this for us? eg:

```tsx
// type helper
type StitchesComponent<C> = TStyledComponentProps<C> & TExtractVariantProps<C>;

// usage
function PageContainer(props: StitchesComponent<typeof Container>) {}
```

We could sill expose them individually, but I imagine most people would need both types, most of the time, no?

jjenzz:
@hadihallak Something like this should get `React.ComponentProps` working I believe:

```jsx
export interface IStyledComponent<ComponentOrTag extends React.ElementType, Variants, Config extends TConfig> extends React.FC<MergeElementProps<React.ElementType, VariantASProps<Config, Variants>> & {
  as?: React.ElementType;
  css?: TCssWithBreakpoints<Config>;
  className?: string;
  children?: any;
}> {
  /**
   * Props of a styled component
   */
  <As extends React.ElementType = ComponentOrTag>(
    // Merge native props with variant props to prevent props clashing.
    // e.g. some HTML elements have `size` attribute. And when you combine
    // both types (native and variant props) the common props become
    // unusable (in typing-wise)
    props: MergeElementProps<As, VariantASProps<Config, Variants>> & {
      as?: As;
      css?: TCssWithBreakpoints<Config>;
      className?: string;
      children?: any;
    }
  ): any;
  /**
   * Compound Variant typing:
   */
  compoundVariant: (
    compoundVariants: VariantASProps<Config, Variants>,
    possibleValues: TCssWithBreakpoints<Config>
  ) => IStyledComponent<ComponentOrTag, Variants, Config>;

  /**
   * @deprecated
   */
  defaultProps?: never;
}
```

I deprecated the `defaultProps` there too because you'll most likely get TS errors and they're probably not worth solving given that they're deprecated.

hadihallak:
That's Awesome @jjenzz  💯 

I did two changes:
1- Updated ` extends React.FC<MergeElementProps<React.ElementType, VariantASProps<Config, Variants>> ` so that it's using `ComponentOrTag` instead of `React.ElementType` as this was causing the component to lose its original non-variants props
2- Removed the as prop from the type of `React.FC` as it wouldn't have an effect on the props one React.ComponentProps is called

After playing with this more [I found out that the thing with the components getting a more broad typing happened also in styled-components](https://codesandbox.io/s/infallible-bash-4pmdq?file=/src/App.tsx) and they have type util that allows the user to get the proper types. not sure why they haven't used something similar to what you suggested or if there are gotchas we're not aware of 🤔 

I removed the old utils but part of me is thinking that `TExtractVariantProps` might be helpful if you only care about variants and nothing more. let me know what you guys think

peduarte:
Thanks @hadihallak 

Some thoughts on the naming, how about:
- `StitchesComponent` 
- `StitchesVariants`

Usage examples:
```tsx
// extract all props of `Button`, inc `as` and `variants`
function MyButton(props: StitchesComponent<Button>) {}

// only extract `variants`
function MyComponent(props: StitchesVariants<Button>) {}
```

@jonathantneal @jjenzz what do you think?

jonathantneal:
For exposing types, I like these a lot. They look immediately clear in purpose, quickly writable, and structured well for IntelliSense completion.

jjenzz:
@peduarte As far as I understand it, we don't need the `StitchesComponent` type because consumers can do `React.ComponentProps<typeof Button>`. Or do we think it's worth exporting `type StitchesComponent<C = React.ElementType>

jjenzz:
@peduarte As far as I understand it, we don't need the `StitchesComponent` type because consumers can do `React.ComponentProps<typeof Button>`. Or do we think it's worth exporting something like the following as well regardless?

```jsx
export type StitchesComponentProps<C = React.ElementType> = React.ComponentProps<C>
```

peduarte:
@jjenzz I think its a nicer API, a nice to have. But yeah, that should work too 👍

hadihallak:
Updated the PR with a new name for the variantsprops util and a StitchesComponentProps util that deals directly with the internal types instead of React.ComponentProps

peduarte:
Oops, merged too soon and ended up committing to `canary`, my bad 🤦 


Commit here:
https://github.com/modulz/stitches/commit/e64c73e34c426912314ebb9558933c3cce16ef90

Renamed `StitchesComponentProps` to `StitchesProps`

pronevich:
@jjenzz After _defaultProps_ deprecated, how default variant should be applied?

Issue #118, opened 2020-08-20
TS: Improve typing 


hadihallak:
we have a lot of instances where we're casting things as `any` internally. we should improve these throughout the codebase

PR #274, opened 2020-10-11
Reduce the usage of TypeScript any in codebase

andykenward:
Relates to #118 

Have made a start at trying to reduce the uage of TypeScript `any` and castings of typings used in the codebase.

 Started with `createCss` [`cssInstance`](https://github.com/modulz/stitches/compare/canary...andykenward:improve_ts?expand=1#diff-55646d75364ce42cdb84abec1a1531ccR613-R636) in the core package. As this was disconnected from the interface typings of  [TCss](https://github.com/modulz/stitches/compare/canary...andykenward:improve_ts?expand=1#diff-2991521f751aa9cf68976ceef512e19aR405-R426)

I've add a couple of TypeScript [User-Defined Type Guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) to help with `[ATOM]` objects. 
Example of usage [here](https://github.com/modulz/stitches/compare/canary...andykenward:improve_ts?expand=1#diff-55646d75364ce42cdb84abec1a1531ccL420-R427)
- [`isAtom`](https://github.com/modulz/stitches/commit/a747c4a550e4091648ee52cb76ee1d2b27628c0a#diff-9731eac31947a9ee3816b17a76ad0badR275-R285)
- [`isComposedAtom`](https://github.com/modulz/stitches/compare/canary...andykenward:improve_ts?expand=1#diff-9731eac31947a9ee3816b17a76ad0badR287-R292)

Noticed in [composeIntoMap](https://github.com/modulz/stitches/compare/canary...andykenward:improve_ts?expand=1#diff-55646d75364ce42cdb84abec1a1531ccR426-R427) & [cssInstance](https://github.com/modulz/stitches/compare/canary...andykenward:improve_ts?expand=1#diff-55646d75364ce42cdb84abec1a1531ccR623) that it can be a string. Is this correct as you end up with `atoms: (IAtom|string)[]` on a `IComposedAtom` instead of the defined [`atoms:IAtom[]`](https://github.com/modulz/stitches/blob/2b0a1e2e2ca413f1af96f166538809481c634351/packages/core/src/types.ts#L116) ? I've left it as is and added the typings that it can be a string and not just an `IAtom` passed around.

Ran into an issue with [cssInstance.keyframes](https://github.com/modulz/stitches/compare/canary...andykenward:improve_ts?expand=1#diff-2991521f751aa9cf68976ceef512e19aR415) `IKeyframesAtom` usage in a couple of unit tests. As `cssInstance` [parameters](https://github.com/modulz/stitches/compare/canary...andykenward:improve_ts?expand=1#diff-2991521f751aa9cf68976ceef512e19aR407) did not have the typings defined that an object with a `IKeyframesAtom` was being passed.

So added a couple `@ts-ignore` for now until someone can have a look at them to solve the issue.
- [should allow keyframes atom to be used as a direct object value](https://github.com/modulz/stitches/compare/canary...andykenward:improve_ts?expand=1#diff-f212137c77c45e774722b9be132e9dc7R969-R970)
- [should inject styles for animations into sheet](https://github.com/modulz/stitches/compare/canary...andykenward:improve_ts?expand=1#diff-f212137c77c45e774722b9be132e9dc7R988-R989)

Cleaned up some `@ts-ignore` and `// tslint:disable-next-line`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,3 +105,6 @@ dist
 
 # TernJS port file
 .tern-port
+
+# VSCode config folder
+.vscode
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -20,6 +20,8 @@ import {
   getVendorPrefixAndProps,
   hashString,
   specificityProps,
+  isComposedAtom,
+  isAtom,
 } from './utils';
 export * from './types';
 export * from './css-types';
@@ -287,7 +289,6 @@ const toStringCachedAtom = function (this: IAtom | IComposedAtom) {
 const toStringCompose = function (this: IComposedAtom) {
   const className = this.atoms.map((atom) => atom.toString()).join(' ');
   // cache the className on this instance
-  // @ts-ignore
   this._className = className;
   // we only want to enable caching on the client
   // because on the server we want to make sure that the composition is evaluated on each request
@@ -382,7 +383,6 @@ const createThemeToString = (classPrefix: string, variablesSheet: ISheet) =>
   function toString(this: IThemeAtom) {
     const themeClassName = `${classPrefix ? `${classPrefix}-` : ''}theme-${this.name}`;
 
-    // @ts-ignore
     variablesSheet.insertRule(
       `.${themeClassName}{${Object.keys(this.definition).reduce((aggr, tokenType) => {
         // @ts-ignore
@@ -411,20 +411,20 @@ const createKeyframesToString = (sheet: ISheet) =>
     return this.id;
   };
 
-const composeIntoMap = (map: Map<string, IAtom>, atoms: (IAtom | IComposedAtom)[]) => {
+const composeIntoMap = (map: Map<string, IAtom | string>, atoms: (IAtom | IComposedAtom | string)[]) => {
   let i = atoms.length - 1;
   for (; i >= 0; i--) {
     const item = atoms[i];
     // atoms can be undefined, null, false or '' using ternary like
     // expressions with the properties
-    if (item && item[ATOM] && 'atoms' in item) {
+    if (isComposedAtom(item)) {
       composeIntoMap(map, item.atoms);
-    } else if (item && item[ATOM]) {
-      if (!map.has((item as IAtom).id)) {
-        map.set((item as IAtom).id, item as IAtom);
+    } else if (isAtom(item)) {
+      if (!map.has(item.id)) {
+        map.set(item.id, item);
       }
     } else if (item) {
-      map.set((item as unknown) as string, item as IAtom);
+      map.set(item, item);
     }
   }
 };
@@ -444,7 +444,7 @@ export const createCss = <T extends TConfig>(
   env: Window | null = typeof window === 'undefined' || window?.Deno ? null : window
 ): TCss<T> => {
   // pre-checked config to avoid checking these all the time
-  // tslint:disable-next-line
+  // tslint:disable-next-line: prefer-object-spread
   const config: TConfig<true> = Object.assign({ tokens: {}, utils: {}, breakpoints: {} }, _config);
   // prefill with empty token groups
   tokenTypes.forEach((tokenType) => (config.tokens[tokenType] = config.tokens[tokenType] || {}));
@@ -491,10 +491,9 @@ export const createCss = <T extends TConfig>(
 
   const { tags, sheets } = createSheets(env, config.breakpoints);
   const preInjectedRules = new Set<string>();
-  // tslint:disable-next-line
   for (const tag of tags) {
     ((tag.textContent || '').match(/\/\*\X\*\/.*?\/\*\X\*\//g) || []).forEach((rule) => {
-      // tslint:disable-next-line
+      // tslint:disable-next-line: prefer-template
       preInjectedRules.add('.' + cleanSSRClass(rule));
     });
   }
@@ -505,7 +504,7 @@ export const createCss = <T extends TConfig>(
 
   let themeToString = createThemeToString(classPrefix, sheets.__variables__);
   let keyframesToString = createKeyframesToString(sheets.__keyframes__);
-  const compose = (...atoms: IAtom[]): IComposedAtom => {
+  const compose = (...atoms: (IAtom | string)[]): IComposedAtom => {
     const map = new Map<string, IAtom>();
     composeIntoMap(map, atoms);
     return {
@@ -521,7 +520,7 @@ export const createCss = <T extends TConfig>(
     selectorString: string,
     inlineMediaQueries: string[],
     isGlobal?: boolean
-  ) => {
+  ): IAtom => {
     // generate id used for specificity check
     // two atoms are considered equal in regard to there specificity if the id is equal
     const inlineMediasAsString = inlineMediaQueries ? inlineMediaQueries.join('') : '';
@@ -535,7 +534,7 @@ export const createCss = <T extends TConfig>(
     if (atomCache.has(uid)) {
       // check if this has a breakpoint based media query
       if (inlineMediasAsString.match(/@media.*\((min|max)?.*(width|height).*\)/)) {
-        // tslint:disable-next-line
+        // tslint:disable-next-line: no-console
         console.warn(
           `The property "${cssProp}" with media query ${inlineMediasAsString} can cause a specificity issue. You should create a breakpoint`
         );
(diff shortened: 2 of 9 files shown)


peduarte:
@andykenward thanks for this <3 @hadihallak has been doing a lot of work around the typings lately so he'll be reviewing this with you. Cheers!

hadihallak (commented):
@andykenward 
Thank you so much for submitting this PR. It looks amazing and I'm super excited to merge this in 🙏 
Overall looks good and only requires some modifications for the things you mentioned which I'm
 leaving feedback for below:

> Noticed in composeIntoMap & cssInstance that it can be a string. Is this correct as you end up with atoms: (IAtom|string)[] on a IComposedAtom instead of the defined atoms:IAtom[] ?

I believe that your assumption is correct however, I think we should fix it by modifying all atom types so that they are normal types instead of interfaces and so that they're types as a string in addition to the main type since they have a .toString() method which basically means that an atom is both a string and whatever the atom is typed as

so IAtom would look something like this:
```ts
// Notice string type in the next line:
export type IAtom = string & {
  id: string;
  cssHyphenProp: string;
  value: string | IKeyframesAtom;
  selector: string;
  breakpoint: string;
  _isGlobal?: boolean;
  inlineMediaQueries: string[];
  _className?: string;
  toString: (this: IAtom) => string;
  [ATOM]: true;
};
```

> Ran into an issue with cssInstance.keyframes IKeyframesAtom usage in a couple of unit tests. As cssInstance parameters did not have the typings defined that an object with a IKeyframesAtom was being passed.

> So added a couple @ts-ignore for now until someone can have a look at them to solve the issue.

I noticed that you modified `TCss` so that it accepts `IKeyframesAtom`. this would work on the upper level but not recursively. I think if you modify `IKeyframesAtom` so that it's also typed as a string like how we did it above, then the types should work correctly

------------

Other than that, this looks perfect 💯 , and again, thank you so much for this PR

Let me know if you have any questions or need help with anything 🙌

andykenward:
@hadihallak So an `IAtom` can be a `string & {}` not `string | {}`?

I dont think `string & {}` would be possible with TypeScript, as it would expect `toString()` to be called to make the object a string.

Also [`createAtom`](https://github.com/modulz/stitches/blob/54da81770402f3394fcd1e5eb4ae88933319e1b9/packages/core/src/index.ts#L517-L570) returns an IAtom as an object. TypeScript complains when `IAtom` is `string & {}`.
 
I understand that we are overriding an object default toString() method. Which JavaScript automatically calls when the object is to be represented as a text value or when an object is referred to in a manner in which a string is expected. 

But I am not sure at the moment how TypeScript deals with this automatic call in JavaScript.

peduarte:
Just chipping in and cc'ing @christianalfoni 

I think I remember Christian saying that `string & {}` was a TS "trick" to speed things up. Could be totally wrong here though.

hadihallak:
@andykenward Yes I believe that this should work as both of the types intersect so if some other type requires one over the other it should not be an issue since we're combining them (I've done a similar change here to allow components to be used as object keys https://github.com/modulz/stitches/pull/280)

> Also createAtom returns an IAtom as an object. TypeScript complains when IAtom is string & {}.

I think in this instance it would be ok to cast the type to `IAtom` because technically our atom is a string when used in the correct context but we can't make typescript infer this kind of usage correctly

Other than the `createAtom` do you see us having any issues with typing atoms as a string? personally, i think it solves a few problems regarding the usage like with the IKeyframesAtom you mentioned.

----------

> I think I remember Christian saying that string & {} was a TS "trick" to speed things up. Could be totally wrong here though.

@peduarte 
I think you're talking about adding `(string & {})` to a union type to prevent typescript from trying to optimize your type
```ts
// typescript will try and act smart to optimize the type so it infers
// the below type to `number | string` and we lose autocomplete for `'hi' | 'there'`
type One = 'hi' | 'there' | number | string 

// ---------

// Replacing `string` with `(string & {})` above will prevent typescript optimizations
// as now we're tricking it so that the below type will correctly preserve autocomplete for `hi` and `there`
type One = 'hi' | 'there' | number | (string & {}) 
```

where @andykenward (please keep me honest) is referring to the type of the atom in his comment 
`string & {}` -> `string & {this is actually the current type of IAtom}`

andykenward:
The issue I have with doing `string & {}` is that an object is not a string and that string is not an object.
As we aren't converting that string back into an object? What happens once it's an actual string?

Also if you try to access properties of an actual string in JavaScript e.g `.toUpperCase()` etc. It will error at runtime in JavaScript when you do this on an actual object. TypeScript will let you write this code and won't complain. But in fact when you run the JavaScript it will error as it isn't an actual string but an object.

I feel like it will make using TypeScript in the codebase a bit redundant. You will always have to check for example if IAtom is a string or object in JavaScript. TypeScript won't help you with this, which is kind of the point of using TypeScript.

It's kind of the issue with `"1"` being a string or a number. We can convert this between a string and a number. And number has a `toString`. But in TypeScript `"1"` it is still a string until you convert it to a number. And same with `1` is a number until you convert it to a string. I think I am making sense???

Wouldn't doing `string | {}` make maintaining the codebase easier in TypeScript. TypeScript will let us know when things are getting converted though out the codebase.

Sorry if I am stepping out of line with continuing the discussion about this. I know I've just started working on this. So I don't really have the whole scope of the project.

I've been trying to find other projects that do something similar with an object `toString` and doing `string & {}` as the typing. Not found one yet.

andykenward:
Perhaps a Conditional Type would be better. Had a quick try of this.

```
export type StringOrObject<T> = T extends string ? string : T;

export type IKeyframesAtom = StringOrObject<{
  id: string;
  _cssRuleString?: string;
  toString: (this: IKeyframesAtom) => string;
  [ATOM]: true;
}>;

export type IAtom = StringOrObject<{
  id: string;
  cssHyphenProp: string;
  value: IKeyframesAtom;
  selector: string;
  breakpoint: string;
  _isGlobal?: boolean;
  inlineMediaQueries: string[];
  _className?: string;
  toString: (this: IAtom) => string;
  [ATOM]: true;
}>;

```

andykenward:
Spent a few more hours on this. Going to close this pull request. I don't recommend you use anything in it.

The more I look into what's currently typed against an `IAtom` in the code base the more its seems too not be the case. And it just opens a huge can of worms.

I would suggest that you start to add code comments to part some knowledge. And if you add `any` in then leave code comments to give some context for when typings get implemented.

Cheers

hadihallak:
> The issue I have with doing string & {} is that an object is not a string and that string is not an object.
As we aren't converting that string back into an object? What happens once it's an actual string?

I totally get that but weirdly enough when you type something as a string and an object, typescript would be ok if you type a string as that type but wouldn't be ok with you typing an object as that type

```ts
type Test = string & {}

// this works:
const value: Test = ''

// this doesn't and will error:
const value2: Test = {}
```


> Also if you try to access properties of an actual string in JavaScript e.g .toUpperCase() etc. It will error at runtime in JavaScript when you do this on an actual object. TypeScript will let you write this code and won't complain. But in fact, when you run the JavaScript it will error as it isn't an actual string but an object.

this is a very valid point, thanks for bringing that up 💯 



> I feel like it will make using TypeScript in the codebase a bit redundant. You will always have to check for example if IAtom is a string or object in JavaScript. TypeScript won't help you with this, which is kind of the point of using TypeScript.

> It's kind of the issue with "1" being a string or a number. We can convert this between a string and a number. And number has a toString. But in TypeScript "1" it is still a string until you convert it to a number. And same with 1 is a number until you convert it to a string. I think I am making sense???

> Wouldn't doing string | {} make maintaining the codebase easier in TypeScript. TypeScript will let us know when things are getting converted though out the codebase.

I think it would still be the case with IAtom | string, you would need to check for the type before doing any work so by typing it as a combined type of string & {} we no longer need to do the above but as you mentioned, this is gonna bring up other issues



> Sorry if I am stepping out of line with continuing the discussion about this. I know I've just started working on this. So I don't really have the whole scope of the project.

Not at all man, This has been incredibly helpful and you're bringing up a lot of important points. will definitely use your work to dial in the types a bit more because as you've mentioned, the types aren't always reflecting the actual type of the used values and this is making things hard to maintain




> I've been trying to find other projects that do something similar with an object toString and doing string & {} as the typing. Not found one yet.

I believe cssinjs libraries like styled-components use a similar technique to allow components to be used as an object key without converting it to string




> The more I look into what's currently typed against an IAtom in the code base the more its seems too not be the case. And it just opens a huge can of worms.

I think you're referring to things like the composeIntoMap function how it returns an array of atoms where in reality it might actually return atoms and strings. yeah these things need ironing out for sure 👍🏻




Thank you so much @andykenward  for all of this great work and debugging 🙏  
I know that working on typescript-related issues and especially at this scale where we have large types files with some of them being wrong isn't fun at all so I'm really grateful you took the time to give us all of this valuable info.

will be using all of the info you presented us to improve the typing of Stitches

PR #275, opened 2020-10-12
Update warnings

peduarte:
Some updates to our `console.warn`'s

- Warning about inline media queries: don't display it in prod and improve the description
- Warning about dynamic `css` prop: removed. Its raising too many questions and we don't have enough data for it. We'll consider re-adding when benchmarks are done (#261)

### DIFF
--- a/packages/core/tests/index.test.ts
+++ b/packages/core/tests/index.test.ts
@@ -1416,9 +1416,7 @@ describe('createCss: mixed(SSR & Client)', () => {
       [mediaString]: { color: 'red' },
     }).toString();
     // tslint:disable-next-line
-    expect(console.warn).toHaveBeenCalledWith(
-      `The property "color" with media query ${mediaString} can cause a specificity issue. You should create a breakpoint`
-    );
+    expect(console.warn).toHaveBeenCalled();
   });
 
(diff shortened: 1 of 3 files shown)


jonathantneal (approved):
This is a quality improvement. Thank you! 🚢

packages/core/tests/index.test.ts
```diff
@@ -1416,9 +1416,7 @@ describe('createCss: mixed(SSR & Client)', () => {
       [mediaString]: { color: 'red' },
     }).toString();
     // tslint:disable-next-line
-    expect(console.warn).toHaveBeenCalledWith(
-      `The property "color" with media query ${mediaString} can cause a specificity issue. You should create a breakpoint`
-    );
+    expect(console.warn).toHaveBeenCalled();
```
hadihallak, line 1419: I think we should check the message here just so that this doesn't catch some other random warning by a mistake

hadihallak (commented):
Looks good 💯 

Just added one small note.

packages/core/tests/index.test.ts
peduarte: Done!

PR #280, opened 2020-10-14
Make components usable as object keys

hadihallak:


packages/react/src/index.tsx
```diff
@@ -81,6 +81,13 @@ type MergeElementProps<As extends React.ElementType, Props extends object = {}>
  * 1. Props of a styled component
  * 2. The compoundVariants function typings
  */
+
+/**
+ * Merges tha passed type with type string
+ * Used to make component behave as strings when they're in style object keys.
```
peduarte, line 87: do you mean "useful"?

Issue #234, opened 2020-09-21
Improve typing of `createStyled`


peduarte:
We need to ensure that only expected keys can be passed to `createStyled` (and `createTokens`)

From @andykenward:

I've been looking at the source code for stitches and i think it might the usage of TypeScript generics on some of the functions.
e.g
```
export const createTokens = <T extends ITokensDefinition>(tokens: T) => {
  return tokens;
};
```

```
export const createStyled = <Config extends TConfig>(
  config: Config
):
```

These will only error if there arent any keys in the object that are defined in the interface.

So for createTokens

```
const tokens = createTokens({
      bob: {},
      colors: {
        RED: 'tomato',
      },
    });
```
is allowed

But if you dont havecolors or any known interface key from ITokensDefinitions then it will error.
for example

```
const tokens = createTokens({
      bob: {},
      colsors: {
        RED: 'tomato',
      },
    });
```

or
```
const tokens = createTokens({
      bob: {},
    });
```

Will both error.

hadihallak:
Spent a lot of time on this and tried a lot of different ways but couldn't get it to work given our use case
right now, the only way I could see this supported is if typescript allowed us to type unknown properties in an object to a type that doesn't intersect with known ones which I think I've seen people discuss previously in one of the issues in typescript so this might eventually be possible.

I'm leaving this minimum repo codesandobx for people to test against so that maybe someone is able to achieve what we're looking for https://codesandbox.io/s/relaxed-noether-ke6ui?file=/src/App.tsx

PR #294, opened 2020-10-24
Typescript public api bug fixes and speed improvements

### LINKED ISSUE #187: TS autocomplete for inline style props is a bit slow
# Bug report

## Describe the bug

Maybe this is more of a clarification than it is a bug. When adding styles to a `styled` component and referencing tokens from the config file, autocomplete has been great and fast.

However, when attempting to override the current styles by using the inline `css` prop, autocomplete is a bit slow. This behavior also occurs on non-style props as well (i.e. `onClick` and other event handlers). 

## To Reproduce

1. Create styled component using the `styled` function
2. Add CSS property referencing token in config file
3. With newly created styled component, override previous styles with `css` prop
4. Notice slow autocomplete on Styled component (on newly created CRA project)

## Expected behavior

Expected fast or same autocomplete speed when referencing tokens through both `styled` function and inline `css` props

## Screen recording

stiches-autocomplete-slow

## System information

- OS: macOS Catalina - v10.15.6
- Version of Node.js: v14.4.0

### LINKED ISSUE #221: Expose a TS Type for style abstractions
We need to expose a TS Type for abstracting CSS Properties.

For example:

```jsx
const sharedStyles: StitchesCss = {
  color: '$purple500',
  '&:hover': {
    color: '$purple700'
  },
  bp1: {
    fontSize: '$4'
  }
}

const Comp1 =  styled('div', {
  ...sharedStyles,
  padding: "$5"
})

const Comp2 =  styled('div', {
  ...sharedStyles,
  position: "absolute"
})
```

hadihallak:
Almost a full rewrite of the types in core + improved types for the react styled package
- A of speed Improvements everywhere including the CSS prop - fixes #187
- Simplified the types in the core package a lot and removed unnecessarily inference 
- Improve typing of the styled object styles - now hovering over a CSS property will show you the possible values where previously this would only show the currently set value
- Improve typing of `createStyled` - fixes #234
- Expose a TS utility to type stitches CSS `StitchesCSS` which will allow you to selectively choose the structure of your CSS by configuring breakpoints, utils, tokens, nested, strict - fixes #221
- Utils inside the config will now be typed to the config (minus the utils) instead of being typed as any

In terms of testing, we need to test that the above stuff is working and that non of the current functionality was broken in the process as a lot of things got re-worked

I'm also thinking that we could delay merging this until next week as typescript 4.1 RC will be landing next week and we could use some of the new features to improve types even more


peduarte:
Is this ready for review? 

Can you add the _Closes_ keyword to close the issues that this is fixing so I can test accordingly?

Here's a link with all the typescript tagged issues: https://github.com/modulz/stitches/issues?q=is%3Aopen+is%3Aissue+label%3Atypescript

hadihallak:
It was until I found a bug with the variants API typings.
will continue working on the fix tomorrow then will re-request review

jonathantneal:
This is looking amazing!

package.json
```diff
@@ -32,7 +32,7 @@
     "tslint-config-prettier": "^1.18.0",
     "tslint-plugin-prettier": "^2.3.0",
     "tslint": "^6.1.1",
-    "typescript": "^3.8.3"
+    "typescript": "^4.0.3"
```
peduarte, line 35: Does Stitches depend on Typescript v4+?

Asking for docs

packages/core/src/types.ts
```diff
@@ -125,297 +35,177 @@ export interface IKeyframesAtom {
...
+  borderTopStyle: 'borderStyles';
+  borderRightStyle: 'borderStyles';
+  borderBottomStyle: 'borderStyles';
+  borderLeftStyle: 'borderStyles';
+  borderRadius: 'radii';
+  borderTopLeftRadius: 'radii';
+  borderTopRightRadius: 'radii';
+  borderBottomRightRadius: 'radii';
+  borderBottomLeftRadius: 'radii';
+  /** Update syntax when ts 4.1 is release */
```
peduarte, line 146: what to does comment mean?

Also, typo nit: `released`

peduarte:
@hadihallak Thanks for all this, proper effort! 👏

I've been testing this in this [codesandbox](https://codesandbox.io/s/stitches-ci-cra-forked-6rwqu), and below are some comments

#### Error using numbers as CSS Property values

I'm unable to use numbers as value (screenshot below).

CleanShot 2020-11-02 at 10 57 54@2x


---

#### Typescript performance when extending components

In this example we expect `Link` to inherit the variants defined in `BaseLink`. 

A couple of things to note:
- The variant name autocompletes eventually, after a long wait
- The value also takes some time, but eventually, it appears in an autocomplete alongside many other incorrect possible values
- Once the variant is applied a typescript error is shown, about overload properties
- After about 1 minute the error goes away
- I tried restarting the browser and typescript server a few times

tstest1

---

#### Typescript performance when autocompleting tokens

Similar to the above, it takes a long time to autocomplete tokens - around 10 to 20 seconds

image

---

### Typescript performance when autocompleting CSS Properties

In this example, I typed in `bac` expecting the autocomplete to suggest `background` related properties. It does autocomplete eventually, after around 15 seconds

image

---

#### Typescript polymorphic props not working as expected

In this example, I change the `Text` component to render as an `a`. At this point I expect to be able to set a `href` prop, but was suggested something else

image

> note: this may be related to performance too, because after saving and waiting, it shows no errors

---

#### createStyled typing

> Improve typing of createStyled - fixes #234

That looks to be working but took about 30 seconds to type check

hadihallak:
@peduarte Thanks for all of this testing 🙏 
I just did a quick test and I think it might be something wrong with the codesandbox build as most things are working for me locally.
will test it later today and let you know what's going on

hadihallak:
@peduarte There seems to be an issue with how types are getting built so the types work well but not in a built version of stitches.

so the typescript issues you showcased are already fixed if not for the broken build which I'm looking into.

regarding "Typescript polymorphic props not working as expected" I've tested this issue and validated its existence in styled-components as well so I'm guessing this is some typescript limitation that's related to typing the as prop. 
ofc the href prop once written should have the correct typing so you will not be able to use href unless you have the as prop set to "a" for example.

package.json
hadihallak: Yes we will depend on 4.1 now that it's a RC

packages/core/src/types.ts
hadihallak: I believe typesciprt 4.1 RC is out now so we should be able to leverage on https://devblogs.microsoft.com/typescript/announcing-typescript-4-1-beta/#template-literal-types
to improve the types for similar properties

hadihallak:
@peduarte This should be ready now again (i hope 😄)
Fixed more issues and added tons of tests to assert typescript types so now our subsequent releases shouldn't break the types very often especially in the core package.

Will be adding more tests later when I work on updating the types to be in line with the new features released in TS 4.1 but this should be good to go

LexSwed:
>In this example, I change the Text component to render as an a. At this point I expect to be able to set a href prop, but was suggested something else

There are some [resources](https://react-typescript-cheatsheet.netlify.app/docs/advanced/patterns_by_usecase#polymorphic-components-eg-with-as-props) and as far as I remember, `@emotion/react` didn't include polymorphic component props in v11 exactly because of the reason mentioned [here](https://stackoverflow.com/a/58612425), it slows down the performance significantly.

jaredLunde:
Wanted to drop in with a few links in case you haven't seen them because the TypeScript checking performance and intellisense performance are huge problems in my project right now since adding stitches. 
https://github.com/chakra-ui/chakra-ui/issues/1582
https://github.com/chakra-ui/chakra-ui/pull/1627

This PR is still using `JSX.IntrinsicElements` in a few places and it seems like that is perhaps not a great idea.

steveruizok:
> Wanted to drop in with a few links in case you haven't seen them because the TypeScript checking performance and intellisense performance are huge problems in my project right now since adding stitches.
> [chakra-ui/chakra-ui#1582](https://github.com/chakra-ui/chakra-ui/issues/1582)
> [chakra-ui/chakra-ui#1627](https://github.com/chakra-ui/chakra-ui/pull/1627)
> 
> This PR is still using `JSX.IntrinsicElements` in a few places and it seems like that is perhaps not a great idea.

I've been in the same boat, unfortunately, and finally decided to pull the plug on the typings by casting my configuration to `any`.

```ts
const { styled, css } = createStyled({
  ... 
}) as any
```

You'll lose all of the typings from the Stitches configuration but your CPU will thank you. The typings are quite complex for Stitches and I wouldn't want to be the one to debug them, but I'm certain that there's a problem there.

hadihallak:
I'm aware of the issue with `JSX.IntrinsicElements` but thankfully I don't believe that this version of the typing (in this PR) is susceptible to it despite the fact that it uses it in a couple of places

I did a lot of research around this performance issue a few months ago and what it boiled down to was that the hit to performance comes to pushing keys of `JSX.IntrinsicElements` into `React.ComponentPropsWithRef` since it's a very large union. 

so, instead of pushing the full union, we use Distributive conditional types to narrow the type to its singular value and then we use that inside `React.ComponentPropsWithRef`so that it never gets the large union that causes the slowdown

Please feel free to check this code sandbox and let me know if you encounter any performance issues when running it
https://codesandbox.io/s/stitches-ci-cra-5hdmh?file=/src/App.tsx

we've been on the quiet side regarding the progress as we've been re-writing stitches with improved performance and even better typings which will be released soon so please keep us updated  in regards to any performance or type improvements you would like to see and we will be sure to include them in the next release

PR #297, opened 2020-10-26
Update bug_report.md

vitordino:
the *System information* section was still pointing at Next.js version instead of stitches 🙈


peduarte (approved):

peduarte:
That's perfect, thank you 👍️

PR #316, opened 2020-12-03
Restore defaultProps keeping the deprecation message

elkevinwolf:
Restore the typing for defaultProps because it was giving TypeScript errors, without removing the deprecation notice.

Issue reproduction:
- https://codesandbox.io/s/broken-architecture-ojw1w?file=/src/App.tsx:301-302
- Open [this file](https://github.com/modulz/stitches/blob/1e777795e6de1edc06a4502817745183bffea239/packages%2Freact%2Ftests%2Findex.test.tsx#L197) locally and TypeScript will complain﻿

### DIFF
--- a/packages/react/src/index.tsx
+++ b/packages/react/src/index.tsx
@@ -130,8 +130,10 @@ export interface IStyledComponent<
 
   /**
    * @deprecated
+   * React's defaultProps will be deprecated for Functional Components.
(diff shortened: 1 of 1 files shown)


jonathantneal:
I’m afraid I need to close this because it predates the beta and canary. We can revisit it as we begin work on the revamped TS.

PR #322, opened 2020-12-18
Ab.type.migration

hadihallak:
Types for core package.

the types are a bit wild in order to cover some edge cases so we can go over them and how they work a few days before the release 

React package types are coming later today


jonathantneal (approved):

hadihallak:
@jonathantneal just checked the build and it's failing because in some areas we're using unknown scales for the theme scale `color` instead of `colors`

and the other instances are things that need casting to any because the react types aren't in yet

PR #324, opened 2020-12-18
Initial React types

hadihallak:


hadihallak:
Will be testing these later to make sure they match the behavior of our polymorphic package but other than that this should be enough to get things moving

PR #330, opened 2021-01-12
Add debugging className based on `showFriendlyClassnames` config

danilowoz:
I've been using Stitches in a personal project, and I've struggled to debug the HTML output and its classNames, as there is no way to know which HTML-tag represents my component, expects by React debug but I can't check out the CSS there. 

So in order to be able to recognize the component on the DOM tree, I've added the `displayName`, which is set in the `StitchesComponent` too, to the main `classNames` based on the `config.showFriendlyClassnames` values. 

Previous output:
```html
<div class="_initial_c_vfarC" />
``` 

Current output using `BodyWrapper` component as an example:
```html
<div class="styled(BodyWrapper) _initial_c_vfarC" />
```

The downside here is that this className is completely useless for styling purpose, but I guess that `showFriendlyClassnames`'s value must be conditional based in the current environment.

```js
const { styled, css } = createStyled({
  showFriendlyClassnames: process.env.NODE_ENV !== "production"
  ...
})
```

Side note: `showFriendlyClassnames` is missing in the documentation, if you don't have any objection about it, I can take care at this later.

### DIFF
--- a/packages/react/src/index.tsx
+++ b/packages/react/src/index.tsx
@@ -232,6 +232,13 @@ export const createStyled = <Config extends TConfig>(
 
     const stitchesComponentId = `scid-${hashString(JSON.stringify(baseAndVariantStyles))}`;
 
+    const componentDisplayName =
+      typeof currentAs === 'string'
+        ? `styled(${currentAs})`
+        : Component && Component.displayName
+        ? `styled(${Component.displayName})`
+        : `styled(Component\)`;
+
     const StitchesComponent = React.forwardRef((props: any, ref: React.Ref<Element>) => {
       const compositions = [baseStyles];
 
@@ -286,7 +293,12 @@ export const createStyled = <Config extends TConfig>(
 
       // By default we don't stringify the classname (composition), so that
       // the children Stitches component is responsible for the final composition
-      let className = css(stitchesComponentId, ...compositions, props.className);
+      let className = css(
+        stitchesComponentId,
+        ...compositions,
+        props.className,
+        config.showFriendlyClassnames ? componentDisplayName : undefined
+      );
 
       // If we're not wrapping a Stitches component,
       // we ensure the classname is stringified
@@ -305,12 +317,7 @@ export const createStyled = <Config extends TConfig>(
 
     (StitchesComponent as any).__isStitchesComponent = true;
 
-    StitchesComponent.displayName =
-      typeof currentAs === 'string'
-        ? `styled(${currentAs})`
-        : Component && Component.displayName
-        ? `styled(${Component.displayName})`
-        : `styled(Component\)`;
+    StitchesComponent.displayName = componentDisplayName;
 
     StitchesComponent.toString = () => `.${stitchesComponentId}`;
 
--- a/packages/react/tests/__snapshots__/index.test.tsx.snap
+++ b/packages/react/tests/__snapshots__/index.test.tsx.snap
@@ -8,7 +8,7 @@ exports[`styled Breakpoints work in variants 1`] = `
     "._breakpointTwo_h_hiyRrM {height: 20px;}"
   ],
   "props": {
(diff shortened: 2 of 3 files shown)


peduarte:
Hey, thanks! Have you tried using this https://www.npmjs.com/package/babel-plugin-transform-stitches-display-name ?

danilowoz:
Hey @peduarte
I haven't used this before, but I just tried and it seems it doesn't solve the same issue I have had. Actually, it attaches the variable name to the display name, but not to the `className` - which is my goal. 

The output of this babel plugin is: 
```js
let Wrapper = Object.assign({}, styled('div', {}), {
  displayName: "Wrapper"
});
```
Then the React-DOM tree will have this display name on its component. 

I think both approaches would be perfect, in order to make the classname/displayname easier to recognize which component I'm looking for. Does it make sense? 

---

I'd like to add saying that as Stitches has a configuration called `showFriendlyClassnames` I think it's totally expected to have a clue or a way to debug and recognize the component in the HTML-tree.

Thanks again

peduarte:
@danilowoz we'll take your suggestion into account for the update of Stitches we're working on. I dont think we can show this without a babel plugin though, so your suggestion of passing a `showFriendlyClassnames` flag to the config will likely not be sufficient. 

This is because the name of the component is assigned to a variable by the user, this is outside stitches, eg:
```jsx
const Button = styled('button', styles)
```

Stitches doesnt know you've called this `Button` when it created the `button` element for you.

danilowoz:
@peduarte I agree, there is no way to get the variable name as display name without a babel plugin. But the best we can do here is extend the same behavior it has on `StitchesComponent` to the className.

Anyway, thanks and let me know about the futures changes.

peduarte:
@danilowoz hey, in the new direction we wont be using atomic classes anymore. the new classes are now component-driven and display hints for variants and override, which I think is a great improvement in the debugging experience. 

To show friendly class names of the component name, it will need to be done with babel, otherwise it's as helpful as the `element` name. 

Anyway, I'll have to close this PR since the codebase has changed a lot, so its no longer comparible.

Thanks! <3

danilowoz:
No problem @peduarte.
I don't know if you've addressed this question in another please, but what is the reason for leaving the atomic approach? As I'm a fan of this kind of approach, I'd like to understand the context of this decision. 

Thanks

peduarte:
@danilowoz 
We'll prob write a blog post expanding on why we switched from atomic CSS to object-oriented CSS. In a nutshell, the main problems with atomic CSS were:
- Lots and lots of specificity issues
- Forced us into making strong opinions within the lib, like decide the specificity order of pseudo-classes
- Poor debugging experience 
- Bloated bundle size to accommodate for lots of edge cases
- Inability to create global rules and global declarations like @font-face, @import, etc. 

In the beta version, we have:
- Reduced bundle size from 8kb to less than 3kb
- Not opinionated - you just write CSS and expect the CSS you wrote
- Excellent debugging experience, inc class names "hints" for variants and overrides
- Fully implemented token parsing
- Supports all features of CSS,  like @font-face, @import, @supports, etc. 

Stitches was never about it being atomic. It was about the DX, the API, and performance. All of those are now better than ever. It went from 8kb to less than 3kb. The performance is incredible - we'll have some benchmarks on this.

ruijdacd:
@peduarte
Do you reckon the API will change much from beta to the first release? I'm currently migrating a component library entirely built with SCSS and have been trying out several other libs over the last 3 months or so, and Stitches is without a doubt the most beneficial one for our use case.

Having an understanding if the lib (and its API) would change drastically, would help us make a decision on if we should take the risk and go forward with Stitches, or if we should be more mindful and for the foreseeable future go on a different route with perhaps Emotion.

peduarte:
@ruijdacd Not much, this is the WIP migration guide https://stitches-site-git-beta.modulz-deploys.com/blog/migrating-from-alpha-to-beta

PR #331, opened 2021-01-26
seperate & Improve types on jn.oo

hadihallak:
1- Types are now being applied using JSDocs so that they're less restrictive for you when working on the internals of the library
2- Did some improvements on the types themselvs


hadihallak:
weird... I pulled before branching yet there seem to be large conflicts so I will close this and create a new PR with a new branch as it's easier than to resolve the conflicts

hadihallak:
Ok I think I got them resolved so I'm re-opening

hadihallak:
@jonathantneal 
Some quick notes:
- Initially, I used JSDocs to keep types separate from the internal implementation but seeing that you are already using them I switched to using casting in the entry ts file to separate them while also making sure this works with preconstruct 
environments
- I updated properties to utils in the types. please let me know if there are any new changes that need to be reflected in the types
- I added back  the CSS-types file since it's needed for the types to work correctly
- converted defaultThemeMap back to ts so that we can leverage on it in the types as casting isn't supported in JSDocs and this map is needed to construct the correct types

peduarte:
🔥 🔥 🔥

PR #337, opened 2021-02-09
Abdul.types.4

hadihallak:
New:
- Public types now live in a types folder and are written manually
- Updated the `when` syntax to the new version
- Polymorphism is improved and tested in many different ways
- Improve TS props resolution with polymorphism so things like the CSS prop on typing props, in general, should be fast to resolve and type check 

Known issues that I’m debugging and will create a new PR for:
- Some component props aren’t autocompleting in vscode although they type check correctly
- Extending other stitches components is causing a TS panic `styled(OtherStitchesComponent, {})`


jonathantneal (approved):
Thank you! 🙏

I’m bringing this in, fixing anything that’s breaking, and removing stuff like the yarn error log.

❤️

PR #345, opened 2021-02-12
Expose config and update utils

hadihallak:
just exposed the config as Pedro and I discussed
also updated the types for the utils in the config


jonathantneal (approved):
🙏 Thank you! Merge when you’re ready 👍

PR #369, opened 2021-02-15
Remove unused types in `css-types.ts`

yenbekbay:
This PR removes unused CSS property types. Unused types are problematic because they cause type-checking to fail if you have the `noUnusedLocals` flag enabled in your `tsconfig.json`.

I also looked at https://github.com/frenic/csstype for reference and these types don't seem to be present there anymore either.


peduarte:
Thanks! /cc @hadihallak

jonathantneal (approved):

PR #400, opened 2021-02-16
Fix auto complete with polymorphism 

hadihallak:
This PR fixes autocompletion in the best way possible with the current limitation of how TS only shows auto-completion for the first overload.

this PR works by changing the interface that we create when we call `styled(T, {})` based on the value of T

**Two main interfaces used here:**
- so if T is a react component then the resulting typed interface is `StitchesComponentWithAutoCompleteForReactComponents`
- and if T is a Stitches component or a JSX intrinsic element then the resulting typed  interface will be `StitchesComponentWithAutoCompleteForJSXElements`

**How events will be typed:**
- events with the as prop will be typed if the as prop was a JSX intrinsic element
- events will be any if the as prop is a component or if the stitches component was extending another non-stitches component using `styled(Component,{})`

**how autocompletion works:**
- for normal usages without the as prop autocompletion should work completely and everything should be typed
- when there is an as prop and for components using `StitchesComponentWithAutoCompleteForReactComponents`, autocomplete will work for as props that have a react component as a value
- when there is an as prop and for components using  `StitchesComponentWithAutoCompleteForJSXElements`, autocomplete will work for as props that have a jsx element as a value

**Type checking:**
- should always work and error correctly regardless of the stitches interface used to construct the type

### DIFF
--- a/packages/react/tests/types.test.tsx
+++ b/packages/react/tests/types.test.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react'
-import createStyled, { StitchesCss, StitchesVariants } from '../types/index.d'
+import createStyled, { StitchesCss, StitchesVariants, StitchesComponentWithAutoCompleteForReactComponents, StitchesComponentWithAutoCompleteForJSXElements } from '../types/index.d'
 import * as SeparatorPrimitive from '@radix-ui/react-separator'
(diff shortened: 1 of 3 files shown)


jonathantneal (approved):
Thanks for walking me through this. 👍

PR #426, opened 2021-02-19
Fix not being able to compose css calls 

### LINKED ISSUE #420: [0.1.0-canary.8] problem with simple composition
# Bug report

## Describe the bug

Typescript throws error when I try to create simple composition with `css` function.

```
Index signature is missing in type 'IStyledRule<{}, TConditions, {}, {}, CSSPropertiesToTokenScale>'.
```

## To Reproduce

https://codesandbox.io/s/frosty-wave-dr8gs?file=/src/index.ts

hadihallak:
This resolves #420

### DIFF
--- a/packages/core/tests/types.test.ts
+++ b/packages/core/tests/types.test.ts
@@ -86,8 +86,6 @@ const PotatoButton = css.css({
 	},
 	compoundVariants: [
 		{
-			//cool
-
 			variant: 'blue',
 			css: {
 				backgroundColor: '$gray200',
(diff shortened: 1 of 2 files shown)


packages/core/types/index.d.ts
```diff
@@ -373,7 +369,7 @@ export interface TStyledSheet<A extends TConditions = {}, B extends TTheme = {},
 								}
 							}
 						)
-						| Record<string, InternalCSS<A, B, C, ThemeMap>>
+						| Record<string, InternalCSS<A, B, C, ThemeMap> | string | number>
```
jonathantneal, line 372: Oh I do prefer this. 👍

packages/core/tests/types.test.ts
```diff
@@ -84,18 +84,49 @@ const PotatoButton = css.css({
...
+	],
+})
+
+const two = css.css(PotatoButton, {
+	$$max: '2px',
+	width: '$$max',
+	variants: {
+		variant: {
+			green: {
+				width: '$$max',
```
jonathantneal, line 103: May we add tests for overriding `$$max` within here? Similarly within any place `when` is supported?

jonathantneal (commented):
This looks great so far! I don’t see tests for deeper uses of custom tokens (like `$$max`) so I’m going to test it a bit further before approving or merging. 👍

Issue #431, opened 2021-02-21
[0.1.0-canary.12] Compound variants do not apply to composed components


arempe93:
# Bug report

## Describe the bug

If a component is composed (eg: `const NewComponent = styled(StitchesComponent, { ... })`, compound variants will not be applied.

## To Reproduce

https://codesandbox.io/s/dazzling-proskuriakova-xihgq?file=/src/App.tsx

Apparent in inspector, all variant class names but no `[xxxxx]--comp` class names are applied.

Comment/uncomment Line 84 to see that single variants will style correctly.

## Expected behavior

Compound variants should apply to composed components

## System information

- Version of Stitches: 0.1.0-canary.12

PR #437, opened 2021-02-24
fix: util recusion and prefixing

### LINKED ISSUE #382: Utils goes into infinite loop when returning its own property in an object
If we have a util that returns its own property in an object, it goes into an infinite loop.

for example:
```jsx
content: config => value => ({
  content: `"${value}"`
})
```

Maybe it was introduced after adding `config`.

jonathantneal:
|Sandbox| Source |
|--|--|
|[Stitches CI: CRA](https://codesandbox.io/s/stitches-ci-cra-forked-cuqmz)|Configuration|
|[Stitches CI: Next.js](https://codesandbox.io/s/stitches-ci-nextjs-forked-wmio6)|Configuration|

<br />

## Fix: Recursive Utility Functions

This PR updates the internal `getComputedStyle` function to fix an issue with recursion in utility properties.

```js
createCss({
  utils: {
    appearance: (config) => (value) => ({ WebkitAppearance: value, appearance: value })
  }
})
```

Resolves #382

---

<br />

## Feature: Utility Submodules

This PR adds utility prop submodules for content, logical properties, & autoprefixes.

**Utility props that allow loosely formatted `content` properties:**
```js
import utils from "@stitches/react/utils/content"

createCss({ utils })

// now `{ content: "Hello world" }` just works™
// instead of having to write `{ content: "'Hello world'" }` with extra quotes
```

**Utility props that apply polyfills for logical properties:**
```js
import utils from "@stitches/react/utils/logical-properties"

createCss({ utils })

// now `{ marginBlock, marginInline, paddingBlock, paddingInline, inset, ...etc }` all work
```

**Utility props that apply prefixes to modern prefix-required properties:**
```js
import utils from "@stitches/react/utils/prefixes"

createCss({ utils })

// now `{ appearance, backgroundClip, tabSize, userSelect, ...etc }` all work
```

**All utility props combined:**
```js
import utils from "@stitches/react/utils"

createCss({ utils })
```

---

<br />

## Feature: Custom Units

This PR adds support for Houdini-style Custom Units.

**What you had to wrote before:**
```js
{ width: "calc(3 * var(--lh))" }
```

**What you get to write now:**
```js
{ width: "3--lh" }
```


peduarte (commented):
Amazing work as always 👏 

Overall I like the idea of relying on utils to vendor-prefix properties. I also think using utils to polyfill certain unavailable CSS properties, like `inset`, is a good idea.

Im on the fence on whether Stitches should ship with these though. If I am building something for a modern browser and I want to use `inset` property, I dont want to then see `top`, `right`, etc when I inspect it. 

Let's jump on a call to discuss this further! Thanks

peduarte (approved):

arempe93:
hey @jonathantneal -- tried this branch in my project since it should fix #431, discovered a bug with `defaultVariants` and composed components

https://codesandbox.io/s/default-variants-bug-11qbb
`TypeError: Cannot read property 'undefined' of undefined`

I tracked it down to this line here:

https://github.com/modulz/stitches/blob/be2734e7e1ed7ddf00865c8d28855025beabd70b/packages/core/src/index.js#L243

because `variantProps[defaultVariantName]` could be `undefined`. So may be a simple fix to guard that, but I'm not aware of the potential second order effects with all the changes.

PR #438, opened 2021-02-24
Add note about Node version in CONTRIBUTING

arempe93:
While following the CONTRIBUTING guide I found I was unable to run the build or test scripts on Node v12 LTS. Installing Node v14 allowed the scripts to run.


jonathantneal:
🎩 Thank you, @arempe93!

Issue #449, opened 2021-03-01
Consider replacing `when` with "something better"


jonathantneal:
**Is your feature request related to a problem? Please describe.**

We would like the ability to intersperse aliased media queries between properties. This isn’t possible without inline media queries, which are not aliased. We would like something like [Custom Media Queries](https://www.w3.org/TR/mediaqueries-5/#custom-mq).

```pcss
@custom-media --smallViewport (width <= 30em);

.component {
  /* styles for a component */

  @media (--smallViewport) {
    /* styles for a component in a small viewport */
  }

  /* more styles for a component */
}
```

Currently, conditional aliases are only accessible within the `when` property. The `when` property serves as a hook into accessing the `conditions` authored in a Stitches configuration. The `when` property in Stitches is not "_just CSS™_". JavaScript prohibits us from putting multiple `when` properties in a rule, so interspersing then with `when` is a non-starter.

**Describe the solution you'd like**

```pcss
@custom-media --smallViewport (width <= 30em);

.component {
  /* styles for a component */

  @media (--smallViewport) {
    /* styles for a component in a small viewport */
  }

  /* more styles for a component */
}
```

Or something else reflective of real CSS.

**Describe alternatives you've considered**

The following syntaxes have been kicked around:

```ts
{
  "@when (--smallViewport)": {
    ...styles
  }
}
```

```ts
{
  [config.conditions.smallViewport]: {
    ...styles
  }
}
```

```ts
{
  $conditions$smallViewport: {
    ...styles
  }
}
```

**Additional context**

I think the more our API resembles something that could ship natively, the more likely it is so serve people better and longer.

peduarte:
Any API that requires the user `import`'ing something is out of the question IMO.

My favourite is `@when` but typing here will be tricky, if we can nail it, I think thats the winner. It's close to CSS and I think thats a great thing. 

`$` should be reserved for tokens I think

Some other ideas:
```jsx
{
  'when@smallViewport': {
    ...styles
  }
}
```

```jsx
{
  '@smallViewport': {
    ...styles
  }
}
```

jonathantneal:
Here’s what we’ve come up with so far:

---

A **custom when query** is defined with the `@condition` rule:

```
@condition <extension-name> <boolean-condition-list> ;
```

The `<extension-name>` can then be used to represent a `<boolean-condition-list>`.

```
<boolean-condition-list> = <boolean-condition>*
```

```
<boolean-condition> = media() | supports()
```

[...]

PR #450, opened 2021-03-02
Features & Fixes: DOM in core, less DOM writes, N Composition, 

### LINKED ISSUE #397: [stitches/core] `css` should accept `n` compositions
Power!

```jsx
const a = css({ color: "red" });
const b = css({ fontSize: 20 });
const c = css({ fontWeight: 'bold' });
const final = css(a, b, c);
```

### LINKED ISSUE #416: [0.1.0-canary.x] Pass variants as props to wrapped components
# Bug report

## Describe the bug

When extending a stitches styled components variants styles of parent and child are merged but when extending a wrapped component only the style of the child is applied because the variant prop is not passed down to the component and the default value is applied.

## To Reproduce

[codesandbox example](https://codesandbox.io/s/stitches-pass-variants-down-to-wrapped-component-9fz2m)

## Expected behavior

I don't know if this is something you want to support or not but it can be useful.

### LINKED ISSUE #431: [0.1.0-canary.12] Compound variants do not apply to composed components

### LINKED ISSUE #436: Theme function not returning class name when using `css` API
When I add the theme as a classname using the `styled` API, it all works fine, as you can see here: https://codesandbox.io/s/stitches-theme-difference-between-apis-styled-version-gbxek?file=/src/App.js

When I do the exact same using the `css` API, the classname isnt added. I need to explicitly call `theme.className`. As you can see here https://codesandbox.io/s/stitches-theme-difference-between-apis-2ig5i?file=/src/App.js

Theming should work consistently across both APIs

### LINKED ISSUE #440: `@stitches/core` should handle injection/hydration of styles
This will allow core to be used on its own, in a platform agnostic way.

### LINKED ISSUE #441: Variants and css prop styles do not apply
# Bug report

## Describe the bug

`Div` component variants and css prop styles do not apply.

## To Reproduce

https://codesandbox.io/s/stitches-bug-su9fq?file=/src/App.tsx

## Expected behavior

Variant and css prop styles are applied.

## System information

- Version of Stitches: 0.1.0-canary.12

## Additional context

If multiple stitch components have the same style, the component's base CSS rules will be duplicated and will take precedence over the additional CSS rules.

### LINKED ISSUE #448: How to handle `as` prop on `styled(Component)` components
**Is your feature request related to a problem? Please describe.**

What I am about to describe is something I have found confusing with `styled-components` and also exists in `stitches` and we had someone report this as a bug in our Primitives Discord channel over the weekend.

They tried something equivalent to the following:

```jsx
const Primitive = ({ as: Comp = 'div', ...props }) => <Comp data-radix-primitive {...props} />;
const Item = styled(Primitive, { color: "red" });
const ButtonBase = styled("button", { fontSize: "16px" });

render(<Item as={ButtonBase}>Loses `Primitive` attrs and functionality</Item>)
```

_A working example of the above code: https://codesandbox.io/s/xenodochial-oskar-ywq1z?file=/src/App.js:43-221_

When doing this, the expectation was that `Item` would become a `button` with `color: red; fontSize: 16px` styles but also maintain the `data-radix-primitive` attribute and `Primitive` functionality. That is not the case.

The `as` prop will replace the component passed to `styled()` entirely with the component passed in the `as` prop. This tends not to be expected but also seems a bit of an issue for us specifically given that we're advocating this is how people would style Primitives while maintaining the `as` prop for use in their application.

**Describe the solution you'd like**

I'm not raising as a bug because honestly, I'm not sure how this **should** behave. 

It behaves the same with `styled-component`s, but you will see that they also provide a `forwardedAs` prop to achieve what we expect: https://codesandbox.io/s/recursing-zhukovsky-qvkow?file=/src/App.js

My expectation is that it would behave the same as `styled-components` but I've never been keen on the `forwardedAs` prop. It might be the only way though.

**Describe alternatives you've considered**

Hoping the team might have some ideas 🙏

jonathantneal:
This PR would introduce the following changes:

- Fixed issues composing together and passing along variants, compoundVariants, and defaultVariants.
- Reduced how often the DOM is modified without reducing how quickly it is updated.
- Added DOM injection to core.
- Added the ability to create a component extended from an unlimited number of other components.
- Changed the `theme` function to return an object instead of another function.

### PR Goals

- [x] Resolves #441
- [x] Resolves #440
- [x] Resolves #436
- [x] Resolves #431
- [x] Resolves #416
- [x] Resolves #397


peduarte:
🔥

jonathantneal:
I’ve removed #448 from the goals of this PR, because like #449 the issue acknowledges that there isn’t an obvious problem or solution to address, neither inside or outside of Stitches.

peduarte (approved):

arempe93:
hey @jonathantneal -- tried this branch in my project since it should fix #431, discovered a bug with `defaultVariants` and composed components

https://codesandbox.io/s/default-variants-bug-11qbb
`TypeError: Cannot read property 'undefined' of undefined`

I tracked it down to this line here:

https://github.com/modulz/stitches/blob/be2734e7e1ed7ddf00865c8d28855025beabd70b/packages/core/src/index.js#L243

because `variantProps[defaultVariantName]` could be `undefined`. So may be a simple fix to guard that, but I'm not aware of the potential second order effects with all the changes.

jonathantneal:
@arempe93, thank you 🙏! Slipped the fix into #456

PR #459, opened 2021-03-09
fix: change unit-ed property matching regex to exclude line height

### LINKED ISSUE #457: [0.1.0-canary.14] Numeric line height is converted to px
# Bug report

## Describe the bug

I had initially reported this in a comment here https://github.com/modulz/stitches/commit/26e2ab13558270dbdd763520d9b47ac1b1e8eabc#r47572049, but since a release was made I'll add it as a tracked issue.

Since canary.14, any unitless `lineHeight` property

```js
lineHeight: 1.5      // either as a number
lineHeight: "1.5"    // or as a numeric string
```

get converted to px

```css
line-height: 1.5px;
```

## To Reproduce

https://codesandbox.io/s/line-height-bug-cslrw

## Expected behavior

Line height accepts numeric values to match CSS spec.

## System information

- Version of Stitches: 0.1.0-canary.14

arempe93:
Apply the fix to the unit-ed property matching regex suggested by @jonathantneal 

Fixes #457

### DIFF
--- a/packages/core/src/createGetComputedCss.js
+++ b/packages/core/src/createGetComputedCss.js
@@ -6,7 +6,7 @@ import isDeclaration from './isDeclaration.js'
 const captureTokens = /([+-])?((?:\d+(?:\.\d*)?|\.\d+)(?:[Ee][+-]?\d+)?)?(\$|--)([$\w-]+)/g
 
 /** Unit-ed property name matcher. */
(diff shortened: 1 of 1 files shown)


jonathantneal:
If you are still around and want to copy my tests from https://github.com/modulz/stitches/pull/460/files then I will gladly merge yours and we can get your name into the commits of this project (if it isn’t already). 😄

jonathantneal:
I’ll merge this anyway, then mine to get in the tests. Thank you for helping with this issue multiple times.

Issue #462, opened 2021-03-09
[0.1.0-canary.14] Composed forwardRef components do not render correctly


arempe93:
# Bug report

## Describe the bug

If you compose a `forwardRef` component

```ts
const Button = forwardRef((props, ref) => { ... })

const StyledButton = styled(Button, { ... })
```

the React component will not render, and a default `span` will be rendered instead.

## To Reproduce

https://codesandbox.io/s/stitches-composition-bug-mdk50

> Change to stitches canary.12 to see working behavior

## Expected behavior

Composed components that forward ref should render.

## System information

- Version of Stitches: 0.1.0-canary.14

arempe93:
I discovered that in this code

https://github.com/modulz/stitches/blob/4b717fbd4bed9df4c4eb17110d887dac2ba4288e/packages/core/src/index.js#L338-L351

if I add an additional check to the L345 condition it works

```ts
init.$$typeof !== Symbol.for('react.forward_ref')
```

however, considering this is core code, that doesn't seem adequate. something should be done on the react side instead.

there may be a clever solution given that: while you inherit _styles_ from **ALL** composers, you will only ever _render_ the **LAST** composer.

peduarte:
Hey, yep I just noticed that today too. Here's my repro with a Radix UI Component https://codesandbox.io/s/quizzical-faraday-d9wjk?file=/src/App.js

PR #463, opened 2021-03-09
Fix forward ref components not rendering correctly

arempe93:
A proposed fix for #462, moves the "type determination" concern to the react lib and out of core.

All the tests pass and it works in my medium-sized project, but I'm not 100% sure of all the implications. Feedback appreciated.

### DIFF
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -330,23 +330,19 @@ const createCss = (init) => {
 	}
 
 	const css = (...inits) => {
-		let type
 		let composers = []
 		let composer
 		let defaultVariants = create(null)
 
 		for (const init of inits) {
 			if ($$composers in Object(init)) {
-				type = init.type || type
 				for (const composer of init[$$composers]) {
 					composers.push(composer)
 					assign(defaultVariants, composer.defaultVariants)
 				}
-			} else if (init && typeof init === 'object' && !('type' in init)) {
+			} else if (init && typeof init === 'object') {
 				composers.push((composer = createComposer(init)))
 				assign(defaultVariants, composer.defaultVariants)
-			} else {
-				type = ('type' in Object(init) ? init.type : init) || type
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


packages/core/src/index.js
```diff
@@ -330,23 +330,19 @@ const createCss = (init) => {
...
 				for (const composer of init[$$composers]) {
 					composers.push(composer)
 					assign(defaultVariants, composer.defaultVariants)
 				}
-			} else if (init && typeof init === 'object' && !('type' in init)) {
+			} else if (init && typeof init === 'object') {
 				composers.push((composer = createComposer(init)))
 				assign(defaultVariants, composer.defaultVariants)
-			} else {
-				type = ('type' in Object(init) ? init.type : init) || type
```
jonathantneal, line 349: This is a difficult problem, and I’ve been thinking about it.

This wasn’t here _just_ for React. It was the catch-all for any type-like value, including functions (which can also be components) or anything that didn’t seem like a component.

I think, to really pull this from core, a function similar to this needs to exist in react.

packages/core/src/index.js
```diff
@@ -330,23 +330,19 @@ const createCss = (init) => {
...
 
 		for (const init of inits) {
 			if ($$composers in Object(init)) {
-				type = init.type || type
 				for (const composer of init[$$composers]) {
 					composers.push(composer)
 					assign(defaultVariants, composer.defaultVariants)
 				}
-			} else if (init && typeof init === 'object' && !('type' in init)) {
+			} else if (init && typeof init === 'object') {
```
jonathantneal, line 343: If `type` is pulled, this can become the `else` statement.

```suggestion
 			} else {
 ```

jonathantneal:
Bringing this in, and probably changing it a bit more.

jonathantneal:
☝️ Indeed. But your idea to leave it out of core is the right start. Thus the merge. Thank you.

Issue #466, opened 2021-03-09
[0.1.0-canary.14+] Utils with object parameters are no longer applied


arempe93:
# Bug report

## Describe the bug

If a util takes in an object as a parameter

```ts
{
  simpleGradient: (config) => ({ from, to, deg = 90 }) => ({
    backgroundImage: `linear-gradient(${deg}deg, ${from} 0%, ${to} 100%)`
  })
}
```

it will not be applied. utils that take in strings, numbers, etc will be applied.

## To Reproduce

https://codesandbox.io/s/stitches-util-object-bug-9yb24

> Switch to canary.12 to see working behavior

## Expected behavior

Utils should be able to define objects as input.

## System information

- Version of Stitches: 0.1.0-canary.14

jonathantneal:
I’ve been waiting for someone to ask about this. 😄

As far as I know, the only way to safely support this would be to force all selectors to include `&`. Otherwise, how would stitches know the difference between `simpleGradient` the property and `simpleGradient` the descendant selector?

arempe93:
right, I don't think there is an elegant way to differentiate completely. I'd propose simply giving utils higher priority.

so if we have a situation like this

```ts
// config utils
{
  article: (config) => ({ color = '$textDark', size }) => ({
    color: color,
    padding: size,

    fontSize: '$lg',
    lineHeight: 2
  })
}

// styled({ ... })
{
  width: '100%',

  article: { size: '$3' }
}
```

we use the util because we know `'article' in utils == true`, as opposed to creating a `.sxabc article` selector.

then you have two escape hatches as a developer:

1. rename the utility (eg: add prefix)
2. use an `'& article':` key just in this instance for more specificity of your intention

jonathantneal:
Let’s add the option I mentioned to the mix, even if it may (or may not) be unpopular:

### Requiring selectors use the nesting selector

```js
// config utils
{
  article: (config) => ({ color = '$textDark', size }) => ({
    color: color,
    padding: size,

    fontSize: '$lg',
    lineHeight: 2
  })
}

// styled({ ... })
{
  width: '100%',

  // pass `{ size: '$3' }` to the `article` utility, otherwise to the `article` property
  article: { size: '$3' }

  // create a nested rule from `& article`
  "& article": { size: '$3' }
}
```

dylanklohr:
[...]

PR #469, opened 2021-03-10
Allow utilities to accept objects as arguments

arempe93:
Allows utilities to accept objects as arguments using the method @jonathantneal suggested in #466.

NB: this change **REQUIRES** you to use `&` for nested selectors. ie if you were doing this before:

```ts
styled('div', {
  '> *': { ... }    // this syntax works in all stitches canary versions
}
```

you now must write it as

```ts
styled('div', {
  '& > *': { ... }  // this & syntax is required by this change
}
```

Fixes: #466

### DIFF
--- a/packages/core/src/createGetComputedCss.js
+++ b/packages/core/src/createGetComputedCss.js
@@ -8,6 +8,9 @@ const captureTokens = /([+-])?((?:\d+(?:\.\d*)?|\.\d+)(?:[Ee][+-]?\d+)?)?(\$|--)
 /** Unit-ed property name matcher. */
 const captureUnited = /(art|ding|dth|End|[Gg]ap|eft|[^e]Height|^height|op|[Rr]ight|rgin|[^b]Size|^size|tom|us)$/
 
+/** Nested selector name matcher */
+const captureSelector = /[@&#\.]/
+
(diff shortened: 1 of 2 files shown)


packages/core/src/createGetComputedCss.js
```diff
@@ -90,11 +93,17 @@ const createGetComputedCss = (config) => {
 				/** Whether the current style is a condition (i.e. media or supports query). */
 				const isCondition = name.charCodeAt(0) == 64
 
+				/** Whether the current style is a CSS selector */
+				const isSelector = !selectors.length || captureSelector.test(name)
```
arempe93, line 97: the `!selectors.length` code is to detect `global` styles, where all top level keys are treated as literal CSS selectors

jonathantneal:
I like this. I need a little more time to review it.

You may be interested in checking out this [CodePen](https://codepen.io/jonneal/pen/53f7b9706eaeb82b2bf128858d9314ae). It’s the pure "CSS-in-JS" part of Stitches, which comes in around 0.5kB. It also works as a good starting point for supporting root selectors while still supporting objects as values.

jonathantneal:
Closing as superseded by the work of `@stitches/stringify`. Thank you for hashing this out with me and for your review of that PR.

PR #477, opened 2021-03-12
@stitches/stringify

### LINKED ISSUE #466: [0.1.0-canary.14+] Utils with object parameters are no longer applied

jonathantneal:
## Proposal

This PR would provide the `stringify` function from **Stitches** as its own package — `@stitches/stringify`.

- 📘 [Documentation](https://github.com/modulz/stitches/tree/jn.stringify/packages/stringify)
- 🏟️ [Playground](https://codepen.io/jonneal/pen/53f7b9706eaeb82b2bf128858d9314ae)
- 🛠️ [Resolves #466](https://github.com/modulz/stitches/issues/466)
- 🛠️ [Resolves #469](https://github.com/modulz/stitches/issues/469)

<br />

### What is the `@stitches/stringify` package?

**Stringify** converts a JavaScript object to a CSS string, optionally replacing values if a replacer function is specified.

This function is a mirror of the `JSON.stringify` function, only instead for generating CSS.

<br />

### Why a `@stitches/stringify` package?

**Stitches** includes a very small, highly effective, performant CSS serializer capable of producing nearly-if-not-all forms of CSS. Unfortunately, the browser does not provide such functionality, and there is no other package yet that can stringify JavaScript objects into this many forms of CSS, and not while also downsampling complex nesting with a high level of accuracy. As a package that does all of this in under 1kB, such functionality may be useful to more developers and in more situations than **Stitches** can intend to provide. This may include runtime-specific CSS libraries, CSSOM enhancements, Node serializers, packaged components, etc.

Therefore, separating this package may provide us an opportunity to strengthen a core component within **stitches** while also helping move the web forward in many new ways.


packages/stringify/src/index.js
```diff
@@ -0,0 +1,134 @@
...
+		/** Object representing the current CSS. */
+		style,
+
+		/** Styled rule selectors representing the current CSS. */
+		selectors,
+	) => {
+		for (let name in style) {
+			/** Whether the current style is a condition (i.e. media or supports query). */
+			const isCondition = name.charCodeAt(0) == 64
+			const isRule = isCondition ? !/^@(charset|import|namespace)$/.test(name) : /&/.test(name) || !selectors.length
```
arempe93, line 36: elegant and flexible way to handle things 🥇

sidenote: pretty sure `name.indexOf('&') != -1` would be much better perf-wise

jonathantneal: I really appreciate the review, especially from you, @arempe93.

I’m pleased to share that code golfing your suggestion to `~name.indexOf('&')` either breaks even or out-performs your `!= -1`, which greatly outperforms `/&/.test` just as you suggested.

Cite: https://jsben.ch/lcvO3

jonathantneal: Resolved in https://github.com/modulz/stitches/pull/477/commits/e2b0cf861262e228cd35a0fafc991e9cbee4b4ee

Testing `includes` yielded similar results as `indexOf`, so I landed there. Please feel free to comment anywhere things seem vague or troublesome.

PR #501, opened 2021-03-22
Feature: Updated insertion method

jonathantneal:
This PR changes the default insertion method to reduce the likelihood of style changes.

- During a push of styles, the update to `<style />` is delayed by one frame, tho _only_ if the tag already has content.
- Each push of new changes will restart the delay like a debouncer.
- If a delay to write modified CSS is fulfilled, then `<style />` will be updated.

### DIFF
--- a/.bin/build-package.js
+++ b/.bin/build-package.js
@@ -26,7 +26,7 @@ const variants = {
 		transform(code, exports) {
 			const iifeExports = []
 			for (const name in exports) iifeExports.push(`${name}:${exports[name]}`)
-			return `${code}globalThis.stitches={${iifeExports.join(',')}}`
+			return `(()=>{${code}globalThis.stitches={${iifeExports.join(',')}}})()`
 		},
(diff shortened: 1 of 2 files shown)


peduarte (approved):

felixakiragreen:
I could be completely wrong, but I _think_ this broke styling when using Vite & HMR. 

It's my fault for using the bleeding edge, no worries! I rolled back to `0.1.0` and it's working again. Just wanted to let you know just in case it was useful info.

jonathantneal:
@felixakiragreen, were you running `0.1.1` or `0.1.2`? There was a bug from another feature in `0.1.1` that we didn’t see until after publish that we squashed in `0.1.2`.

felixakiragreen:
Pretty sure I was running `0.1.2` because I noticed it was just published a few hours earlier.

Good news, working smoothly with `0.1.3`. 🙏 Many thanks!

PR #504, opened 2021-03-23
Feature: Let utilities override polyfills, do not apply both

jonathantneal:
This PR changes how polyfills are applied, so that polyfills can be replaced by utilities, rather than apply the polyfills after utilities. This will prevent issues where user-authored polyfills create duplicate solutions, or at worst infinite vendor prefixes.


peduarte:
Awesome!

peduarte (approved):

PR #505, opened 2021-03-23
0.1.1

jonathantneal:
- Polyfills the vendor-prefixed sizing values `stretch` and `fit-content`.
- Maps the `outline` property to the `color` theme token.
- Run utilities for properties instead of polyfills, instead of running them beforehand.
- Update the DOM even more sparingly.


peduarte (approved):
Let's get it!

PR #511, opened 2021-03-24
Remove `version` from root `package.json`

peduarte:
Also to test that #510 works 😄


jonathantneal (approved):
Yes! I’d prefer to not have a version here. I recall there were complaints whenever a package.json was missing this, but if that’s fixed or ignorable then awesome!

peduarte:
@jonathantneal Yeah I think its fine to remove it from the root one. I checked a few popular monorepos like Next.js and they don't have it either

Issue #357, opened 2021-02-13
Integrating stitches in a lib, with the possibility of this lib being used in an app using stitches.


dbismut:
Hi - First of all let me just say once again that I'm really fond of `stitches`, really feels like a no-compromise styling lib for React.

### The context
We're building a react GUI which main API will be a hook. People will use the hook and a GUI pane will be automatically spawn. The GUI should be fully theme-able. We've been switching from Styled Components to `stitches` because we faced a wall, namely Styled Components recommending to [add `styled-components` as a peer dependency](https://styled-components.com/docs/faqs#marking-styledcomponents-as-external-in-your-package-dependencies).

This resulted in poor DX, we want people to just add our lib and not bother with SC. So `stitched` felt like a great candidate for replacement and so far so good, we're super happy with our move!

### The problem
The lib isn't out and stitches is not widespread yet (I'm sure it will) so the problem didn't surface yet, but I've just tried integrating our lib into a minimal app consuming stitches `createCss` default export and it breaks the lib styling (doesn't trigger errors, but the theme that our lib should create doesn't inject any css vars).

Here is a sandbox. Try uncommenting line 6 and our gui pane lib will disappear.
https://codesandbox.io/s/leva-minimal-forked-4qkm4

So it looks like the `createCss` that we call in our lib doesn't have any effect.

Furthermore, there is a chance that even if that works, our variables would conflict with the ones that a user could create for its app.

### Possible solutions

I'm not sure how you can allow several `createCss` calls in the app. But we can't ask people to use our own `styled/css` export from our theme. The GUI is supposed to be a light, disposable element, and we want it to be dropped in or out easily without hassle.

- Ideally, we could scope the lib to a node. This may be a bit harder than it looks, since the node might not exist until it's actually rendered by React.
- Otherwise, namespacing the css vars would limit conflicting issues.

### Libs used
- @stitches/react 0.1.0-canary-1
- react + react-do 17.0.0 

Thanks!

AndrewPrifer:
Hi, Andrew here, I'm also involved with the aforementioned React GUI.

I don't think we need to worry about namespacing anything, because ideally, we'll be using shadow dom. What we do need is a way to

1. scope the generated stitches variables / styles to the root of our shadow dom
2. create a localized stitches cache, if there is any (not quite sure about the stitches internals)

[...]

PR #536, opened 2021-04-02
add commonHash to core

### LINKED ISSUE #472: CSS custom property namespacing/hashing
## Context

It goes without saying that stitches is a _fantastic_ tool, and admittedly, the issue described in this issue isn't something that is caused by stitches. Rather, it's something that I believe with a tweak to stitches, at a minimum, would allow for stitches to be more easily introduced **existing** codebases as well as support additional complex use-cases. This issue has been created to continue a conversation started on https://github.com/modulz/stitches/issues/357.

## The Problem

CSS custom properties generated by stitches currently take the form of `--<token scale>-<token name>`. When defining a configuration like

```js
const stitchesConfig = createCss({
  theme: {
    colors:{
      primary: '#0063f3'
    }
  }
});
```

a CSS custom property is created like so: `--colors-primary: #0063f3`.

The issue here arises when we introduce stitches with such a config into a package or application that already defines an identically named `--colors-primary` property and a naming-collision occurs, causing either the existing `--color-primary` or our _new_ stitches created `--color-primary` token to be overwriten. The introduction of `insertMethod` in #456 _helps_ to control which definition for `--color-primary` "wins" out in the end, but I don't believe it allows for both existing and the new stitches defined custom properties to co-exist.

## Desired Solution

While style encapsulation/scoping via the shadow DOM would help to avoid the issue, doing so in every environment, isn't always feasible – particularly in larger applications with existing custom properties present.

Ultimately, I believe an ideal solution would allow for any _existing_ CSS custom properties in the environment to co-exist side-by-side with the new tokens created by stitches, without completely placing the burden on consumers to avoid some of this environmental complexity.

## Potential "Fixes"

1. Allowing for the current `prefix` option that's passable to `createCss` to also prefix new css-custom properties.

e.g.

```js
const stitchesConfig = createCss({
  prefix: 'my-thing',

  theme: {
    colors:{
      primary: '#0063f3'
    }
  }
});
```

would produce something like `--my-thing-colors-primary`, `--my-thing--colors-primary`, etc.

2. automatically (or via an opt-in config flag of some sort) provide hashing to new css custom properties to greatly reduce the change of collision.

e.g.

```js
const stitchesConfig = createCss({
  theme: {
    colors:{
      primary: '#0063f3'
    }
  }
});
```

would produce something like `--<hash>-colors-primary`, `--colors-primary-<hash>`, etc.

3. Provide an optional "format" function that can be specified via the config that extends control to consumers about how they would like their classes or CSS custom properties namespaced.

e.g.

```js
const stitchesConfig = createCss({
  format: (name: string ) => `my-thing-${name}`,

  theme: {
    colors:{
      primary: '#0063f3'
    }
  }
});
```

This _feels_ more inline with the sort of control inversion that `insertMethod` provides. If this option was chosen I think it would be completely reasonable to **require** that any function provided to `format` (or whatever better name is come up with) be **completely deterministic**.

---

Again, thank you for all of your awesome work on this library! My hope is that a change like the one described above would allow for consumers to more easily integrate stitches into exsting codebases 😄

dylanklohr:
This PR adds the optional `commonHash` configuration parameter (not attached to the naming 😄) to `createCss` to provide greater opt-in collision avoidance to `@stitches/core` and `@stitches/react`.

When a `commonHash` property is provided to `createCss`, the following have said hash added as a suffix

- `style` element id targeted by the default `insertMethod`
- generated CSS custom properties
- generated class names

I've tested this in several personal apps and packages consuming stitches, and haven't noticed any issues. 

---

- resolves #534 
- closes #472
- related to #357


jonathantneal:
Wow! I really like this PR, and would like to integrate this sort of feature in the next release.

Let me reach out to you about some work going on in `jn.performance.overhaul`, and make sure this idea is not lost in the "overhaul".

PR #538, opened 2021-04-06
Fix: Do not add an undefined className to a DOM Element

### LINKED ISSUE #537: `className={undefined}` adds an `"undefined"` class to the DOM element
# Bug report

## Describe the bug

Adding `className={undefined}` to a Stitches styled component adds an `"undefined"` class to the DOM element.

## To Reproduce

Live reproduction: https://codesandbox.io/s/stitches-classname-undefined-qopjx?file=/src/App.js

## Expected behavior

`className={undefined}` should not add an `"undefined"` class to the DOM element.

## System information

`@stitches/react v0.1.4`

flozia:
Styled elements with `className={undefined}` do not add a class "undefined" to the DOM element anymore.

- resolves #537

### DIFF
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -384,7 +384,7 @@ const createCss = (init) => {
 					}
 				}
 
-				if ('className' in props) {
+				if ('className' in props && props.className) {
 					String(props.className).split(/\s+/).forEach(classNames.add, classNames)
 				}
 
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -384,8 +384,8 @@ const createCss = (init) => {
 					}
 				}
 
-				if ('className' in props && props.className) {
-					String(props.className).split(/\s+/).forEach(classNames.add, classNames)
+				if (typeof props.className === 'string') {
+					props.className.split(/\s+/).forEach(classNames.add, classNames)
 				}
 
 				const classNameSetArray = from(classNames)
--- a/packages/react/tests/component-className-prop.js
+++ b/packages/react/tests/component-className-prop.js
@@ -1,0 +1,33 @@
+import createCss from '../src/index.js'
+
+describe('className prop', () => {
+	test('Renders a DOM Element with a class matching the className prop', () => {
+		const { styled } = createCss()
+
+		const component = styled('div')
+		const className = 'myClassName';
+		const expression = component.render({ className })
+
+		expect(expression.props.className).toBe(className)
+	})
+
+	test('Renders a DOM Element with multiple classes passed as className', () => {
+		const { styled } = createCss()
+
+		const component = styled('div')
+		const className = 'myClassName1 myClassName2 myClassName3';
+		const expression = component.render({ className })
+
+		expect(expression.props.className).toBe(className)
+	})
+
+	test('Renders a DOM Element withoup adding an undefined class', () => {
+		const { styled } = createCss()
+
+		const component = styled('div')
+		const className = undefined;
+		const expression = component.render({ className })
+
+		expect(expression.props.className).toNotBe('undefined')
+	})
+})


DanTsk:
Merge it please, really need it! 😄

peduarte:
Thanks for the contribution 🙏 

New features should ideally be accompanied by a test case, to prevent future regressions.

Would you mind adding one? Thanks!

packages/core/src/index.js
```diff
@@ -384,7 +384,7 @@ const createCss = (init) => {
 					}
 				}
 
-				if ('className' in props) {
+				if ('className' in props && props.className) {
 					String(props.className).split(/\s+/).forEach(classNames.add, classNames)
```
rafgraph, line 388: @peduarte why not just use `typeof className === 'string'`?
```js
if (typeof props.className === 'string') {
  props.className.split(/\s+/).forEach(classNames.add, classNames)
}
```

I'm not sure using `String(/* some truthy value*/)` is what we want...
```js
String(true) // 'true'
String({}) // '[object Object]'
String(() => {}) // '() => {}'
```

rafgraph: If you're using `String(...)` to call the `toString` function like React, then might want to match React and not convert booleans to strings.

Also, TypeScript types doesn't support non-string values for `className` (for both Stitches and React).

peduarte: Thanks.  @jonathantneal will review this soon 🙌

jonathantneal: Yes. Perhaps we should not be passing any `undefined` values to the `props` forwarded to a component.

I think a good solution would be to create an internal `forwardProps` object kind of like this: `forwardProps = { ...props }`, and then loop through the original `props`, but remove variant names and `undefined` values from `forwardProps`. After all processing, `forwardProps` is forwarded as `props` to the component. Also, this way we don’t touch the original `props`.

I will use this PR as inspiration for that change, with credit to @rafgraph and @flozia for helping with in the issue.

rafgraph: Sounds good. I'm not familiar with the internals of Stitches, but I believe at some point you'll need merge the original `className` prop with the Stitches `className` before passing it to the React element.

rafgraph: FWIW React handles `undefined` values for props without issue, and I've found it to be an easy way to [disable props](https://github.com/rafgraph/react-interactive/blob/f11996fe5420d21e500539aaaf32f7c45708d062/src/index.tsx#L737-L750) without the need for addition logic, but maybe this is bad practice.

jonathantneal: > I believe at some point you'll need merge the original className prop with the Stitches className before passing it to the React element.

You are right, which is the scenario this PR addresses. 😄 Thank you for helping me understand the issue more clearly.

The current behavior matches the expectations of `className` and `classList.add`.

```js
document.body.className = undefined

// <body class="undefined">
```

```js
document.body.classList.add(undefined)

// <body class="undefined">
```

Such behavior does not match the expectations of React props, which specifically ignore `undefined`.

I would not expect any non-string to be split as a string. Therefore, I think this PR can be merged, or this PR can be merged with @rafgraph’s suggestion to gate classes using `typeof props.className === 'string'`, and then changing `String(props.className).split(/\s+/)` to `props.className.split(/\s+/)`.

rafgraph: @flozia what do think about changing the PR to...
```js
if (typeof props.className === 'string') {
  props.className.split(/\s+/).forEach(classNames.add, classNames)
}
```

flozia: Sure – changed it. Thank you for the review and suggestions @rafgraph and @jonathantneal. 🙂

jonathantneal (approved):
This makes things better. Thank you, @flozia!

peduarte (approved):
Thank you 🙏

PR #540, opened 2021-04-08
Fix: made it possible to use ThemeToken instances as css property values

wellguimaraes:
The recommended way to use tokens is nice, but it's not refactoring friendly. If we make it possible to use ThemeToken instances as css prop values, we can still use tokens in a nice way and refactoring works better.

```typescript
import { createCss } from "@stitches/react"

const { styled, theme } = createCss({
  theme: {
    colors: {
      coral: "#f5533d"
    }
  }
})

const Root = styled('div', {
  color: theme.colors.coral
})
```

### DIFF
--- a/packages/stringify/src/index.js
+++ b/packages/stringify/src/index.js
@@ -32,7 +32,8 @@ export const stringify = (
 				}
 
 				const isAtRuleLike = name.charCodeAt(0) === 64
-				const isObjectLike = data === Object(data) && !('length' in data)
+				const isThemeToken = data.constructor.name === 'ThemeToken'
+				const isObjectLike = data === Object(data) && !('length' in data) && !isThemeToken
 
 				if (isObjectLike) {
 					if (used.has(selectors)) {
--- a/packages/stringify/tests/index.js
+++ b/packages/stringify/tests/index.js
@@ -121,4 +121,23 @@ describe('stringify()', () => {
 			'}',
 		)
 	})
+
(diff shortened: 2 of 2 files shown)


jonathantneal:
@wellguimaraes, thank you for identifying and working on this issue.

The current check being shipped is insufficient, and while I like what this does, I think we can patch this in a more permanent way.

For instance, something like this:

```js
isObjectLike = !(
  // the ! means that checks within () are inverted 🤔, so
  // the check is for non-objects 🤨, meaning
  // the non-non-object is an object 🤪.

  // a non-plain-object is not an object; or,
  typeof data !== 'object' ||

  // a non-plain-object is null (because, for historical reasons, null is a type of object); or,
  data === null ||

  // a non-plain-object is an object with a non-Object toString
  data.toString !== Object.prototype.toString
)
```

wellguimaraes:
> data.toString !== Object.prototype.toString

Good alternative @jonathantneal! Thanks!

Even tho, for the sake of less cognitive effort to read I'd rather go with:
```js
isObjectLike = (
  data !== null &&
  typeof data === 'object' &&
  data.toString === Object.prototype.toString
)
```

jonathantneal:
Thanks, @wellguimaraes, and thanks for raising this issue.

Sometimes my brain works funny, and I did implement it in a more straight-forward manner. 😄  https://github.com/modulz/stitches/pull/544/files#diff-01a44dbec706787ef8f2552647f95d5c906fae7cf97c237a148bb734f0b5c8deR37

PR #542, opened 2021-04-08
Fix: Add default className when initialized without CSS prop

### LINKED ISSUE #533: Selector Target a Stitches component a little weird
# Bug report

## Describe the bug

When I do this 
```
[`& ${Box}`]: { display: "none" },
```

it's not working 
when
```
const Box = styled("div");
```
but it's working when
```
const Box = styled("div", { });
```

## To Reproduce

Check this sandbox https://codesandbox.io/s/stitches-demo-with-lee-forked-k5pgv?file=/src/App.js

change 
```const Box = styled("div");```
to 
```const Box = styled("div",  {});```

then the online gray button will display none

## Expected behavior

I'm not sure whether it's expected or not, I think it's easier to write the stitches component like this 
```const Box = styled("div");```
when it doesn't have any custom style

### PR DESCRIPTION
**Pull request by flozia**
When creating a Stitches component without a CSS prop, it is now created with the default className `sx03kze`.

Not adding the default class to the DOM element can lead to confusion. For example when targetting the component with a selector in the styles of a parent.

```
const Label = styled('span');

  [`& ${Label}`]: {
    color: 'lavenderblush'
  }
});
```

- resolves #533

### DIFF
--- a/packages/react/src/index.js
+++ b/packages/react/src/index.js
@@ -16,7 +16,9 @@ const createCss = (init) => {
 			...inits
 		) => {
 			const defaultType = inits.map((init) => (Object(init).type ? init.type : init)).find((init) => init) || 'span'
-			const composition = sheet.css(...inits.filter((init) => $$composers in Object(init) || (init && typeof init === 'object' && !init.$$typeof)))
+			const initsFiltered = inits.filter((init) => $$composers in Object(init) || (init && typeof init === 'object' && !init.$$typeof));
+			const initWithoutStyles = String(defaultType) === defaultType && !initsFiltered.length
+			const composition = sheet.css(...initsFiltered, initWithoutStyles && {})
 
 			/** Returns a React element. */
 			return Object.setPrototypeOf(
--- a/packages/react/tests/component-css-prop.js
+++ b/packages/react/tests/component-css-prop.js
@@ -53,4 +53,16 @@ describe('React Component with CSS prop', () => {
 			'}',
 		)
 	})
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: packages/react/src/index.js [hunk 1]
```diff
@@ -16,7 +16,9 @@ const createCss = (init) => {
 			...inits
 		) => {
 			const defaultType = inits.map((init) => (Object(init).type ? init.type : init)).find((init) => init) || 'span'
-			const composition = sheet.css(...inits.filter((init) => $$composers in Object(init) || (init && typeof init === 'object' && !init.$$typeof)))
+			const initsFiltered = inits.filter((init) => $$composers in Object(init) || (init && typeof init === 'object' && !init.$$typeof));
+			const initWithoutStyles = String(defaultType) === defaultType && !initsFiltered.length
```
flozia, line 20: Check if the Stitches component is supposed to compose a DOM element.

packages/react/src/index.js
```diff
@@ -16,7 +16,9 @@ const createCss = (init) => {
...
+			const composition = sheet.css(...initsFiltered, initWithoutStyles && {})
```
flozia, line 21: Init with empty styles in case no CSS prop was passed.

jonathantneal:
This is a good patch. I will give this a second glance, and this may be good to add as a patch before the `jn.performance-overhaul` work is integrated, which I’d love for you to review. It will guarantee no regressions here.

Thank you for always including ❤️ new tests!

PR #550, opened 2021-04-11
fix: pass `initConfig` to `insertionMethod`

fmal:
`defaultInsertionMethod` checks for `insertionMethod` string to determine if it should append or prepend styles in the head: https://github.com/modulz/stitches/blob/e0a99d32005657a839d94784dab795413eb50eac/packages/core/src/defaultInsertionMethod.js#L12. However the `config` object doesn't have the `insertionMethod` prop, so it's impossible to overwrite the default `prepend` behavior with `append`. The solution is to pass `initConfig`.

### DIFF
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -34,7 +34,7 @@ const createCss = (initConfig) => {
 	/** Prefix added before all generated class names. */
 	const prefix = config.prefix = initConfig.prefix || 'sx'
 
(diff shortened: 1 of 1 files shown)


jonathantneal (approved):

fmal:
@jonathantneal thanks for reviewing! Would really appreciate if you could merge this and https://github.com/modulz/stitches/pull/551 and make a patch release `0.1.6` if that's possible.

jonathantneal:
Yes, a patch release will come this morning. I want to be sure that we want to pass `initConfig` and not mix `insertionMethod` into `config` instead.

PR #556, opened 2021-04-13
fix: support non-undefined className prop values

### LINKED ISSUE #555: `styled`  doesn't accept styles via `className` prop
# Bug report

## Describe the bug

Components created with `styled` doesn't accept styles via `className` prop.

## To Reproduce

https://codesandbox.io/s/sleepy-dubinsky-04gio?file=/src/App.js

## System information

- OS: Windows
- Browser: latest Chrome
- Version of Stitches: @stitches/react@0.1.6

jonathantneal:
This PR fixes an issue where non-string `className` prop values were ignored.

Resolves #555

### DIFF
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -399,8 +399,8 @@ const createCss = (initConfig) => {
 					}
 				}
 
-				if (typeof props.className === 'string') {
-					props.className.split(/\s+/).forEach(classNames.add, classNames)
+				if (props.className !== undefined) {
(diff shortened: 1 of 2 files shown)


jonathantneal:
Confirmed that it works. Yay! 🎉

rafgraph:
@jonathantneal this was my bad for pushing for `typeof props.className === 'string'` in #538, it seems like calling `toString()` on objects is important as per the example in #555.

Though I'm still not sure about converting `boolean`, `function`, `null` to strings in `className` since React doesn't do this.

PR #569, opened 2021-04-19
Add `key: null` to React element object

rafgraph:
This will prevent React from thinking the key value is `undefined`, which I believe is a valid key (in React dev tools it show up as `key="undefined"` on every `styled` component).

### DIFF
--- a/packages/react/src/index.js
+++ b/packages/react/src/index.js
@@ -33,7 +33,7 @@ const createCss = (init) => {
 						} = composition(initProps)
 
 						/** React element. */
-						return { constructor: undefined, $$typeof: $$typeofElement, props, ref, type, __v: 0 }
+						return { constructor: undefined, $$typeof: $$typeofElement, props, ref, type, key: null, __v: 0 }
 					},
 					$$typeof: $$typeofForward,
 					[$$composers]: composition[$$composers],


jonathantneal:
Hey @rafgraph, I am unable to see what you are describing in [React Developer Tools](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi). Would you raise an issue with a reproducible snippet?

To be sure it wasn't something in new React released, I also tested this with React/ReactDOM 17.

screenshot of component details of a stitches component in react developer tools

rafgraph:
Hi @jonathantneal, I can put something together.

There's a chance you might be looking at the element that's created from the [`render` component](https://github.com/modulz/stitches/blob/canary/packages/react/src/index.js#L24) (React will call `createElement()` on the `render` function/component since it's part of a forwarded ref object). The `key="undefined"` shows up on the [element that's returned](https://github.com/modulz/stitches/blob/canary/packages/react/src/index.js#L36) from the `render` component.

In React dev tools the key `value` usually shows up just above where your image is cut off (as well as `key="value"` in the component tree). Also, something that I've found confusing is dev tools hides intrinsic elements by default, so if you're looking at the rendering of `styled('a')` (as opposed to `styled(MyComp)`), you won't see the element with `key="undefined"` by default.

rafgraph:
Here's a screen shot from the Stitches site.

Screen Shot 2021-04-22 at 8 58 46 PM

rafgraph:
Opened issue #579 to track this.

rafgraph:
Closing in favor of #580

PR #571, opened 2021-04-19
Feature: check SSR text before changes

### LINKED ISSUE #523: [0.1.3] FOUC after upgrading from alpha to beta
# Bug report

## Describe the bug

I'm noticing a prominent Flash Of Unstyled Content after migrating to `v0.1.x`

## Screenshots

Before in alpha:

alpha

After in beta:

beta

## Additional Context

We are using CRA and not relying on SSR

jonathantneal:
This PR changes how Stitches hydrates `<style>` tags, so that SSR text is never removed from a sheet, and so that pre-existing styles are never added to the stylesheet. This change also removes the debouncer, which had introduced delayed renderings.

Resolves #523

### DIFF
--- a/packages/core/src/defaultInsertionMethod.js
+++ b/packages/core/src/defaultInsertionMethod.js
@@ -1,50 +1,21 @@
 import { assign } from './Object.js'
 
 export default (init) => {
-	let isSsrCssRemoved = false
-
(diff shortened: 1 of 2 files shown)


jonathantneal:
Closing in favor of a new strategy.

PR #572, opened 2021-04-21
feat: add backdrop-filter polyfill

jemmyphan:
since safari need `-webkit` prefix


peduarte:
@jonathantneal lets see if we can sneak this in too

PR #576, opened 2021-04-22
Fix: Ignore the default prefix on tokens

jonathantneal:
This PR separates "prefixing" from "token prefixing". At this time, Stitches assigns a default `prefix`, which is required to generate valid CSS classes. Authors don’t mind this, but they do mind once the default `prefix` gets assigned to tokens. We are solving this another way — the right way — but as a _patch_ this code creates a **forked prefix just for tokens**, ignoring the default `"sx"` prefix.

This **does not** remove the ability to assign custom prefixes to tokens. The change is limited to the default `"sx"` prefix, and it is only a small, temporary patch.

### DIFF
--- a/packages/core/src/createStringify.js
+++ b/packages/core/src/createStringify.js
@@ -40,6 +40,15 @@ export const createStringify = (config) => {
 
 	const memo = {}
 
+	// TODO: PLEASE REMOVE THIS
+	//       This separates "prefixing" from "token prefixing". At this time,
+	//       Stitches assigns a default "prefix", which is required to generate
+	//       valid CSS classes. Authors don’t mind this, but they do mind when
+	//       the default prefix gets assigned to tokens. We are solving this
+	//       another way — the right way — but as a "patch" this code creates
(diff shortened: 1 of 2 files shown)


packages/core/src/createStringify.js
```diff
@@ -40,6 +40,15 @@ export const createStringify = (config) => {
 
 	const memo = {}
 
+	// TODO: PLEASE REMOVE THIS
```
peduarte, line 43: haha yes yes

peduarte (approved):

jonathantneal:
@dylanklohr, I can confirm that this preserves the ability to define a custom prefix that is assigned to tokens. The tests were (minimally) updated to reflect this. However, I will add a note to the original PR description.

peduarte:
@dylanklohr if you explicitly pass a `prefix`, your tokens will be prefixed with it.

Otherwise, no prefix.

dylanklohr:
@peduarte & @jonathantneal , thank you for the information! 😄 🙏

Issue #454, opened 2021-03-04
Feature request: Support for preact/compat


loiacon:
**Is your feature request related to a problem? Please describe.**
If we try to use `preact/compat` with `@stitches/react` it throws an error:
```
DOMException: Failed to execute 'createElement' on 'Document':
The tag name provided ('.sxp3sgw') is not a valid name.
```

**Describe the solution you'd like**
Just work out of the box 📦, as other tools do

peduarte:
We are working on a "stitches/preact" package as well. The react one won't just work with preact because it has some react-specific behaviours for ref forwarding. 

OOC what do you mean "other libraries do"? Does styled components or emotion work with preact out of the box?

loiacon:
Styled Components works without problems using `preact/compat` as alias for `react`.

I created a demo with NextJS + preact/compat + styled-components & stitches for comparison:
https://codesandbox.io/s/nextjs-preact-styled-components-s6ngt

peduarte:
Hmm, interesting, I wonder how they're able to to forward ref and still be compatible with Preact 🤔 /cc @jonathantneal

jonathantneal:
I am going to look into the issue further, to see what the cause is, and if it is indeed related to `forwardRef`. If that’s the case, then I can understand the appeal to release a Preact version.

Fun fact: before `forwardRef` was introduced to support polymorphism in React, Stitches worked with Preact out of the box, without even the need for `preact/compat`.

vitordino:
it worked out of the box (and incredibly great) on `0.0.3-canary.4`, just tested upgrading to `0.1.0-canary.20` and it stopped working.

ch99q:
To my understanding, it may be caused by the creation of an element through `Object.setPrototypeOf` and not `React.createElement` as seen in https://github.com/modulz/stitches/blob/canary/packages/react/src/index.js#L22.

Normally this function would be replaced with Preact's `createElement` but since Stitches just create the result of the ?`createElement` from React, it's only possible to use React and not Preact.

jonathantneal:
Yes. At first, the beta of this library used a regular object to work with React and Preact out-of-the-box, without any compat. However, supporting polymorphism correctly required `React.createRef`, while the change that made it in only upgraded the object `$$typeof` of the object.

In a future release, it may be better to switch to `React.createRef`, since cross-compatibility out-of-the-box is no longer an option.

rafgraph:
[...]

PR #580, opened 2021-04-23
Fix: React.memo and key="undefined"

### LINKED ISSUE #568: `styled(React.memo(MyComp), {})` removes the memoization
# Bug report

## Describe the bug

`styled(React.memo(MyComp), {})` removes the memoization. `MyComp` is rendered as though it is not memoized.

On a semi-related note, `key="undefined"` is showing up in React dev tools on every component that is wrapped in `styled`. I don't think this is material for performance, but it's an easy fix, ~~just add `key: null` to the React element object [here](https://github.com/modulz/stitches/blob/canary/packages/react/src/index.js#L36)~~ PR [here](https://github.com/modulz/stitches/pull/569).

## To Reproduce

https://codesandbox.io/s/stitches-react-memo-bug-s044r?file=/src/App.js

## Expected behavior

A memoized component to be rendered.

## Screenshots

Screen Shot 2021-04-18 at 11 07 00 PM

## System information

- Version of Stitches: `0.1.7`

### LINKED ISSUE #579: `key="undefined"` added to elements created by Stitches, and "rendered by" list missing in dev tools
# Bug report

## Describe the bug

`key="undefined"` is added to elements created by Stitches, and the "rendered by" list is missing in React dev tools.

## To Reproduce

Inspect React dev tools on a site using Stitches. The "rendered by" list will only show up when using a development build of React.

## Expected behavior

There should be no `key` value set on elements created by Stitches (key should be `null` on the element object), and the "rendered by" list should show the component render stack.

## Screenshots

Stitches

## System information

Stitches `0.1.7`

rafgraph:
Hi, this is a PR for #568 and #579.

Fixes #568 - Stitches removes the memoization from components wrapped in `React.memo`, e.g. `styled(React.memo(MyComp))`.

Partially fixes #579 - Stitches creates elements with `key="undefined"` (fixed), and "rendered by" component stack missing in React dev tools (requires change to using `React.createElement(...)`, not included in this PR).

---

### Fix for #568 `styled(React.memo(Component))`

Essentially the issue is that the `styled` function is returning a React component in the form of a forwarded ref object with some extra Stitches properties. One of those extra properties is `type` which is set to the value of the element/component passed into `styled`.

The `type` property on a forwarded ref object is not used by React (it's ignored), but Stitches uses it when composing multiple calls to `styled`, e.g. `const Div = styled('div'); const GreenDiv = styled(Div, { color: 'green' });`, so there's only one forwarded ref object returned no matter how many times `styled` is composed (instead of returning multiple nested forwarded ref objects).

However, the React memo object (returned from `React.memo(Comp)`) does have a `type` property set the value of the component that it's memoizing. When a React memo object was passed into `styled` like `styled(React.memo(Comp))`, Stitches discarded the React memo object thinking the `type` property was added from a previous call to `styled`, when really the React memo object needs to be used as the `type` for the React element returned from the forwarded ref object's `render` function.

I think it's best to leave the `type` property to React and only use it when it's meant to be consumed by React, so I renamed the one on the forwarded ref object to `stitchesType`. This fixes the issue.

---

These forwarded ref object and memo object examples may be helpful as a reference when reviewing this PR.
React source code: [ReactForwardRef.js](https://github.com/facebook/react/blob/master/packages/react/src/ReactForwardRef.js) and [ReactMemo.js](https://github.com/facebook/react/blob/master/packages/react/src/ReactMemo.js).

```js
// React forwarded ref object, returned from React.forwardRef((props, ref) => {...})
const forwardedRefObject = {
  $$typeof: Symbol.for('react.forward_ref'),
  render: (props, ref) => {
    // the render function needs to return something that react can render
    // for example a react element object, like:
    return <div>hello</div>;
    // OR
    return React.createElement('div', { children: 'hello' });
  },
};
```

```js
// React memo object, returned from React.memo(MyComponent, compare)
const memoObject = {
  $$typeof: Symbol.for('react.memo'),
  type: MyComponent,
  compare: null, // null means react does a standard shallow comparison
  // OR use a custom areEqual function
  // note that React.memo sets compare to null on the object
  // if a value is not passed in to React.memo
};
```

---

### Fix for #579 `key="undefined"`

The `key="undefined"` issue is because `React.createElement(...)` explicitly sets `key: null` on the element object it returns if a key prop is not passed to the React component. This is why when Stitches left key undefined on the element object React interpreted it as a literal key value (only `null` means no key), so `key="undefined"` showed up on the element created by Stitches. I added `key: null` to the React element object which fixes this issue.

It's usually recommended to generate a React element object by calling `React.createElement(...)` instead of manually creating it, as that avoids these pitfalls. Also, with using `React.createElement(...)` it adds the `_owner` property (see below), and other dev only properties for a better dev experience.

I think it would be a good idea to switch to using `React.createElement(...)` to create the element object, but I did not make that change in this PR. I don't think avoiding the `React.createElement` function call is providing a performance benefit (the function call is very inexpensive, especially in production), and AFAICT there aren't any other benefits to Stitches from creating the object manually.

React source code for `createElement`: [ReactElement.js](https://github.com/facebook/react/blob/master/packages/react/src/ReactElement.js)
```js
// React element object, returned from React.createElement
// and the new JSX transform
const elementObject = {
  $$typeof: Symbol.for('react.element'),

  // MyComponent, 'div', etc
  type: type,

  // React.createElement() explicitly sets `key: null`
  // if a key prop is not passed to the React component
  key: key,

  // the props that are on the element object are the props
  // passed into the component with key and ref removed
  props: propsWithOutKeyAndRef,
  ref: ref,

  // records the component responsible for creating this element
  _owner: ReactCurrentOwner.current,
  // this comes from Fiber, Stitches can't access it
  // it's used in dev tools to show the "rendered by" list
  // currently all Stitches created elements just have "react-dom"
  // in the dev tools "rendered by" list b/c _owner is missing

  // additional dev only properties, see React source code for more info
  _store,
  _self,
  _source,
};
```

### DIFF
--- a/packages/react/src/index.js
+++ b/packages/react/src/index.js
@@ -33,7 +33,7 @@ const createCss = (init) => {
 					} = composition(initProps)
 
 					/** React element. */
-					return { constructor: undefined, $$typeof: $$typeofElement, props, ref, type, __v: 0 }
+					return { $$typeof: $$typeofElement, key: null, props, ref, type }
 				},
 				$$typeof: $$typeofForward,
 				displayName: 'Stitches',
--- a/packages/react/src/index.js
+++ b/packages/react/src/index.js
@@ -18,7 +18,7 @@ const createCss = (init) => {
 			const defaultType = inits.map((init) => (Object(init).stitchesType ? init.stitchesType : init)).find((init) => init) || 'span'
 			const composition = sheet.css(...inits.filter((init) => $$composers in Object(init) || (init && typeof init === 'object' && !init.$$typeof)))
 
-			/** This is a React component in the form of a forwarded ref object, with a few extra Stitches properties */
+			/** This is a React component in the form of a forwarded ref object, with a few extra Stitches properties. */
 			return {
 				/** The render function on the forwarded ref object returns a React element. */
 				render(
@@ -26,10 +26,9 @@ const createCss = (init) => {
 					initProps,
 					ref,
 				) {
-					// express the component, extracting `props`, `as` & `ref`
+					/** Express the component, extracting `props` & `as`. */
 					const {
 						props: { as: type = defaultType, ...props },
-						...expressedProps // eslint-disable-line no-unused-vars
 					} = composition(initProps)
 
 					/** React element. */
@@ -38,7 +37,7 @@ const createCss = (init) => {
 				$$typeof: $$typeofForward,
 				displayName: 'Stitches',
 
-				/** Below are all Stitches properties */
+				/** Below are all Stitches properties. */
 				[$$composers]: composition[$$composers],
 				[Symbol.toPrimitive]() {
 					return composition.selector
--- a/packages/react/src/index.js
+++ b/packages/react/src/index.js
@@ -10,7 +10,7 @@ const createCss = (init) => {
 	const sheet = createCoreCss(init)
 
 	return assign(sheet, {
-		/** Returns a React forwarded ref object. */
+		/** Returns a React component in the form of a forwarded ref object. */
 		styled: (
 			/** Type of component. */
 			...inits
@@ -18,9 +18,9 @@ const createCss = (init) => {
 			const defaultType = inits.map((init) => (Object(init).stitchesType ? init.stitchesType : init)).find((init) => init) || 'span'
 			const composition = sheet.css(...inits.filter((init) => $$composers in Object(init) || (init && typeof init === 'object' && !init.$$typeof)))
 
-			/** This is a React forwarded ref object, with a few extra Stitches properties */
+			/** This is a React component in the form of a forwarded ref object, with a few extra Stitches properties */
 			return {
-				/** The render function is the React component on the forwarded ref object, it returns a React element. */
+				/** The render function on the forwarded ref object returns a React element. */
 				render(
 					/** Props used to determine the expression of the current component. */
 					initProps,
--- a/packages/react/src/index.js
+++ b/packages/react/src/index.js
@@ -10,49 +10,50 @@ const createCss = (init) => {
 	const sheet = createCoreCss(init)
 
 	return assign(sheet, {
-		/** Returns a React component. */
+		/** Returns a React forwarded ref object. */
 		styled: (
 			/** Type of component. */
 			...inits
 		) => {
-			const defaultType = inits.map((init) => (Object(init).type ? init.type : init)).find((init) => init) || 'span'
+			const defaultType = inits.map((init) => (Object(init).stitchesType ? init.stitchesType : init)).find((init) => init) || 'span'
 			const composition = sheet.css(...inits.filter((init) => $$composers in Object(init) || (init && typeof init === 'object' && !init.$$typeof)))
 
-			/** Returns a React element. */
-			return Object.setPrototypeOf(
-				{
-					render(
-						/** Props used to determine the expression of the current component. */
-						initProps,
-						ref,
-					) {
-						// express the component, extracting `props`, `as` & `ref`
-						const {
-							props: { as: type = defaultType, ...props },
-							...expressedProps // eslint-disable-line no-unused-vars
-						} = composition(initProps)
+			/** This is a React forwarded ref object, with a few extra Stitches properties */
+			return {
+				/** The render function is the React component on the forwarded ref object, it returns a React element. */
+				render(
+					/** Props used to determine the expression of the current component. */
+					initProps,
+					ref,
+				) {
+					// express the component, extracting `props`, `as` & `ref`
+					const {
+						props: { as: type = defaultType, ...props },
+						...expressedProps // eslint-disable-line no-unused-vars
+					} = composition(initProps)
 
-						/** React element. */
-						return { constructor: undefined, $$typeof: $$typeofElement, props, ref, type, __v: 0 }
-					},
-					$$typeof: $$typeofForward,
-					[$$composers]: composition[$$composers],
-					[Symbol.toPrimitive]() {
-						return composition.selector
-					},
-					toString() {
-						return composition.selector
-					},
-					get className() {
-						return composition.className
-					},
-					get selector() {
-						return composition.selector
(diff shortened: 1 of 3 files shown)


packages/react/src/index.js
```diff
@@ -10,49 +10,50 @@ const createCss = (init) => {
...
-					},
-					get className() {
-						return composition.className
-					},
-					get selector() {
-						return composition.selector
-					},
-					type: defaultType,
+					/** React element. */
+					return { $$typeof: $$typeofElement, key: null, props, ref, type }
```
rafgraph, line 36: I would recommend replacing this with `React.createElement(type, Object.assign(props, ref))`, or use JSX like `<Type {...props} ref={ref} />`. This way the `_owner` property can be added to the element object for a better dev experience, and it will avoid bugs like `key="undefined"` (or something worse) should React internals change in the future (which I don't believe would require a semver breaking change as manually creating the element object is not part of the React api).

packages/react/src/index.js
```diff
@@ -10,49 +10,50 @@ const createCss = (init) => {
...
+				/** The render function is the React component on the forwarded ref object, it returns a React element. */
+				render(
+					/** Props used to determine the expression of the current component. */
+					initProps,
+					ref,
+				) {
+					// express the component, extracting `props`, `as` & `ref`
+					const {
+						props: { as: type = defaultType, ...props },
+						...expressedProps // eslint-disable-line no-unused-vars
```
rafgraph, line 32: I left this in but I'm not sure why `...expressedProps` is needed.

packages/react/src/index.js
```diff
@@ -10,49 +10,50 @@ const createCss = (init) => {
...
 	return assign(sheet, {
-		/** Returns a React component. */
+		/** Returns a React forwarded ref object. */
 		styled: (
 			/** Type of component. */
 			...inits
 		) => {
-			const defaultType = inits.map((init) => (Object(init).type ? init.type : init)).find((init) => init) || 'span'
+			const defaultType = inits.map((init) => (Object(init).stitchesType ? init.stitchesType : init)).find((init) => init) || 'span'
 			const composition = sheet.css(...inits.filter((init) => $$composers in Object(init) || (init && typeof init === 'object' && !init.$$typeof)))
```
rafgraph, line 19: This insinuates that `styled` can take more than more than 2 arguments, like `styled('div', { color: 'green' }, { fontWeight: 'bold' })`. I thought this might be legacy as it's not part of the documented API, so I tried simplifying it to just the two arguments, but it failed a couple of tests and I didn't have time to investigate.

packages/react/src/index.js
```diff
@@ -10,49 +10,50 @@ const createCss = (init) => {
...
 
-			/** Returns a React element. */
-			return Object.setPrototypeOf(
```
rafgraph, line 22: The proptotype of the forwarded ref object was previously set `Object(defaultType)`. This is not needed by React on the forwarded ref object, and AFAICT not needed by Stitches. 

`defaultType` is always the element/component passed in to the first `styled` call (when composing multiple styled calls), and is never the forward ref object returned by styled with additional Stitches properties, so I don't think this prototype makes a difference to Stitches when composing styled calls, but I didn't dig into the code behind `sheet.css(...)`. This passed tests and I tested it out in the apps that I have that use Stitches, and nothing popped up. So I think it's safe to delete.

packages/react/src/index.js
```diff
@@ -10,49 +10,50 @@ const createCss = (init) => {
...
-							...expressedProps // eslint-disable-line no-unused-vars
-						} = composition(initProps)
+			/** This is a React forwarded ref object, with a few extra Stitches properties */
+			return {
+				/** The render function is the React component on the forwarded ref object, it returns a React element. */
+				render(
+					/** Props used to determine the expression of the current component. */
+					initProps,
+					ref,
+				) {
```
rafgraph, line 28: It might be a good idea to replace the manually created forwarded ref object with a call to `React.forwardRef` and then add the Stitches properties to the returned object. This would probably be less brittle in case React internals change since manually creating the forwarded ref object is not part of the React api.

```js
const forwardedRefObject = React.forwardRef((initProps, ref) => {...});

// add Stitches properties to forwardedRefObject
forwardedRefObject.stitchesType = defaultType;
// etc...
// would need to use Object.defineProperty for getters

return forwardedRefObject;
```

packages/react/src/index.js
```diff
@@ -10,49 +10,49 @@ const createCss = (init) => {
...
-							...expressedProps // eslint-disable-line no-unused-vars
-						} = composition(initProps)
+			/** This is a React component in the form of a forwarded ref object, with a few extra Stitches properties. */
+			return {
+				/** The render function on the forwarded ref object returns a React element. */
+				render(
+					/** Props used to determine the expression of the current component. */
+					initProps,
+					ref,
+				) {
```
rafgraph, line 28: It might be a good idea to replace the manually created forwarded ref object with a call to `React.forwardRef((initProps, ref) => {...}` and then add the Stitches properties to the returned object. This would probably be less brittle since manually creating the forwarded ref object is not part of the React api.

packages/react/src/index.js
rafgraph: This would probably also help with supporting Preact #454

packages/react/src/index.js
rafgraph: This would probably also help with supporting Preact #454

jonathantneal:
Hey @rafgraph, since we’ve gone back to using `React.forwardRef`, is this superseded by 0.2.x?

rafgraph:
@jonathantneal @peduarte are #568 and #579 fixed with tests added (especially for #568)? Hopefully you found some of the insights from this PR helpful.

jonathantneal:
Both issues this fixed were resolved in 0.2.x, so I’m closing this. Thank you for the test case, and for the excellent write up. I shared a fork of your reproduction in the other issue to confirm it is fixed. Thank you!

PR #601, opened 2021-05-24
0.2.x - use CSSOM

jonathantneal:
This pull request updates the insertion method in Stitches, switching from DOM insertion to CSSOM insertion. It also resolves a number of issues.

- Replaced DOM insertion method with CSSOM insertion method.
- Removed `insertionMethod` API.
- Fixed `backdropFilter` not transforming locally-scoped tokens in `css()`. (# 588)
- Fixed inconsistent use of `prefix`. (# 586)
- Fixed resulting CSS from custom theme / globals not appending to head if parameters were the same. (# 563)
- Fixed performance degradation after migrating from alpha to beta. (# 535)
- Fixed variant props being passed down to DOM elements. (# 532)
- Fixed flash of unstyled content after upgrading from alpha to beta. (# 523)
- Fixed default variants being ignored when responsively applying variants. (# 492)
- Fixed responsive variants not respecting usage order. (# 483)
- Fixed specificity issue when components would share the same style. (# 474)
- Fixed specificity issue with inherited variants & variants leaking as DOM attributes. (# 473)


rafgraph:
It looks like this PR switches to using the `React.createElement` API 👍 so I think it fixes #579

stefanobrambilla:
hi @jonathantneal , i try to use canary but i got 
`error - TypeError: e.replace is not a function
    at f (webpack-internal:///./node_modules/@stitches/react/dist/index.cjs:1:5401)`

jonathantneal:
Hey @stefanobrambilla, I believe this was fixed in a subsequent canary —https://github.com/modulz/stitches/releases/tag/v0.2.0-canary.2

PR #630, opened 2021-06-15
0.2.x — safely wrap css inserts

jonathantneal:
This changes the functionality for inserting new rules into the CSSOM so that they will not throw when an invalid rule is being added.


peduarte (approved):

peduarte:
All tested ✅

PR #639, opened 2021-06-16
0.2.x — Omit catch warning log on invalid rule

### LINKED ISSUE #637: @stitches/react 0.2.1 t.insertRule is not a function
# Bug report

## Describe the bug

I have a warning in my console when I run test with RTL : `t.insertRule is not a function`. I have this warning on all my components with variants.

I have an other warning on Chrome : `Failed to execute 'insertRule' on 'CSSGroupingRule': the rule '.c-PJLV-javOHm-state-unclickable :hover:not("button"){cursor:default}' is invalid and cannot be parsed.`

## To Reproduce

For the first warning, it on all my unit test of UI components which contains variants.

For the second one, the implementation of the style is the following 

```js
export const Row = styled('tr', {
	variants: {
		state: {
			clickable: { ':hover': { cursor: 'pointer' } },
			unclickable: {
				':hover:not("button")': {
					cursor: 'default',
				},
			},
		},
	},
})
```

## Expected behavior

I would like to not have these warnings, or understand why I have them and how to resolve it.

## System information

- macOS
- Chrome
- Stitches/react : 0.2.1

jonathantneal:
This change removes the `console.warn` that can display when a dynamically inserted rule is unknown to the browser. In most scenarios, this is because the selector is unsupported in that particular browser, and tho the fallback is also included in the stylesheet, the unsupported rule will still log an unnecessary warning.

Resolves #637

### DIFF
--- a/packages/core/src/createSheet.js
+++ b/packages/core/src/createSheet.js
@@ -224,8 +224,8 @@ const addApplyToGroup = (/** @type {RuleGroup} */ group) => {
 			groupingRule.insertRule(cssText, index)
 
 			++index
-		} catch (error) {
-			console.warn(error.message)
+		} catch {
+			// do nothing and continue
 		}
 	}
 }


MaximeConan:
Thanks a lot for the reactivity 👍

Issue #643, opened 2021-06-17
`Token#name` or `Token#token`


jonathantneal:
When accessing the properties of a token, this feels intuitive.

```js
const hue = theme.colors.primary

hue.name  // primary
hue.value // black
hue.scale // colors

hue.variable      // --colors-primary
hue.computedValue // var(--colors-primary)
`${hue}`          // var(--colors-primary)
```

If I’m thinking about a _token_, then I think about the **name** of my token, and then the **value** of my token, etc. If I’m thinking about a _hue_, then I think about the **name** of my hue, and then the **value** of my hue, etc.

---

However, in Stitches 0.1.x, this was accessed differently.

```js
const hue = theme.colors.primary

hue.token // primary
```

This seems a little harder to intuit. If I’m thinking about a _token_, then I think about accessing the **token** of my token. Or, if I’m thinking about a _hue_, then I think about accessing the **token** of my hue.

---

While I have a bias for `Token#name`, let me open this for discussion — should it be `Token#name` or `Token#token`?

PR #644, opened 2021-06-17
0.2.x — rollback `token.name` to be `token.token`

jonathantneal:
This change switches `Token#name` back to `Token#token`, as it was in Stitches 0.1.x.

See https://github.com/modulz/stitches/issues/643


peduarte (approved):
Thanks!

PR #658, opened 2021-07-08
Fix issue with default variants not applying during composition

### LINKED ISSUE #652: Unable to override inherited variants
When we compose a Stitches Component that contains variants, we should be able to override its inherited variants. 

```jsx

  variants: {
    variant: {
      primary: {
        color: "red"
      }
    }
  },
  defaultVariants: {
    variant: "primary"
  }
});

const NewButton = styled(Button, {
  variants: {
    variant: {
      primary: {
        color: "blue"
      }
    }
  }
});
```

In the example above, `NewButton` overrides the `primary` variant, inherited from `Button`. 

Since `Button` sets `primary `as the default variant, that's also inherited. So we should expect `NewButton` to be blue, but it isn't. 

Repro: https://codesandbox.io/s/stitches-bug-overriding-inherited-variants-7sron?file=/src/App.tsx

UPDATE: The current workaround is to re-define `defaultVariants`
```diff
const NewButton = styled(Button, {
  variants: {
    variant: {
      primary: {
        color: "blue"
      }
    }
  },
+ defaultVariants: {
+   variant: "primary"
+ }
});
```

jonathantneal:
This PR changes how default variants are spread across compositions, allowing them to spread much more intuitively.

Resolves #652

### DIFF
--- a/packages/core/src/features/css.js
+++ b/packages/core/src/features/css.js
@@ -320,7 +320,8 @@ const getPreparedDataFromComposers = (/** @type {Set<Composer>} */ composers) =>
 		combinedUndefinedVariants.push(...undefinedVariants)
 
 		for (const name in prefilledVariants) {
-			combinedPrefilledVariants[name] = prefilledVariants[name]
+			const data = prefilledVariants[name]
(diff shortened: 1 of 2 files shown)


peduarte (approved):

peduarte:
Nice one. Let's release this along with a few other bits and bobs today.

Issue #675, opened 2021-07-25
Class name stuck to first Stitches-styled component in composition


LucasUnplugged:
# Bug report

## Describe the bug
- If we pass a Stitches-`styled` component as the 1st param to another `styled` expression, the `className` of the parent is used on the child.
- In cases where there are 3+ levels of composition, the `className` of the first component in the chain is used for _all_ composed components.

## To Reproduce

Example: in this [code sandbox](https://codesandbox.io/s/staging-field-4cs36?file=/src/App.tsx), I created a `Box` primitive, and used it for composition in 2 components: `Header` and `Footer`.

I then placed them both inside another component (`Main`), and styled it as such:
```TSX
const Main = styled(Box, {
  background: '$gray500',
  [`& ${Header}`]: {
    background: '$purple600',
    color: 'white',
  },
});
```

## Expected behavior

In that example, only `Header` should be purple with white text. Instead, both `Header` and `Footer` have that styling.

## Screenshots

<img width="900" alt="Screen Shot 2021-07-24 at 8 41 37 PM" src="https://user-images.githubusercontent.com/2473823/126884223-53119b57-b6ab-48b4-ab3e-31b3dfebcd96.png">

## System information

- OS: macOS
- Browser: Chrome 91.0.4472.114 (Official Build) (x86_64)
- Version of Stitches: 0.2.3
- Version of Node.js: 10.20.1, [according to this](https://codesandbox.io/docs/faq#can-i-change-the-node-version-used-in-a-container-sandbox).

## Additional context

The console logs show that the `Box` primitive's `className` is being used on all composed components (see screenshot).

rafgraph:
There's an atomic-like CSS specificity issue (using style objects instead of single rules) that I think is related. It seems like class names are generated based on deep equal (or style equal) for the style object (instead of referential equality), so if there's an object with matching styles someplace else in the app there will be specificity issues.

See below codesandbox for an example. The `div`s and `span`s are completely independent, but because they have deep equal style objects the `span`s render incorrectly.
https://codesandbox.io/s/pensive-microservice-tdb4y?file=/src/App.tsx

peduarte:
@rafgraph Hey, thanks for chiming in. It is a somewhat related issue - but it's different. Applying Lucas' solution won't solve the problem you're demonstrating. [Sandbox using Lucas' fix as dependency](https://codesandbox.io/s/boring-germain-308mj?file=/src/App.tsx)

[...]

PR #676, opened 2021-07-27
#675 Fix className Composition

LucasUnplugged:
# Description

Addresses issue #675: `className` sticks to initial component, when composing Stitches-styled components.

**NOTE 1**: I renamed `initialClassName` to `latestClassName`, to reflect the change in line 316 of `css.js`, but the only functional change is the one to line 316.

**NOTE 2**: I don't have full context on why `initialClassName` was being used for chaining of class names; switching to `latestClassName` definitely addresses the issue above, but I'm not sure if it could have unintended consequences, due to my lack of context here.

### DIFF
--- a/packages/core/src/features/css.js
+++ b/packages/core/src/features/css.js
@@ -147,13 +147,13 @@ const createRenderer = (
 	/** @type {import('../createSheet').SheetGroup} */ sheet
 ) => {
 	const [
-		initialClassName,
+		latestClassName,
 		baseClassNames,
 		prefilledVariants,
 		undefinedVariants
 	] = getPreparedDataFromComposers(composers)
 
-	const selector = `.${initialClassName}`
+	const selector = `.${latestClassName}`
 
 	/** @type {Render} */
 	const render = (props) => {
@@ -254,7 +254,7 @@ const createRenderer = (
 		// apply css property styles
 		if (typeof css === 'object' && css) {
 			/** @type {string} Inline Class Unique Identifier. @see `{COMPOSER_UUID}-i{VARIANT_UUID}-css` */
-			const iClass = `${initialClassName}-i${toHash(css)}-css`
+			const iClass = `${latestClassName}-i${toHash(css)}-css`
 
 			classSet.add(iClass)
 
@@ -285,13 +285,13 @@ const createRenderer = (
 	}
 
 	const toString = () => {
-		if (!sheet.rules.styled.cache.has(initialClassName)) render()
-		return initialClassName
+		if (!sheet.rules.styled.cache.has(latestClassName)) render()
+		return latestClassName
 	}
 
 	return define(render, {
 		type,
-		className: initialClassName,
+		className: latestClassName,
 		selector,
 		[$$composers]: composers,
 		toString,
@@ -301,7 +301,7 @@ const createRenderer = (
 /** Returns useful data that can be known before rendering. */
 const getPreparedDataFromComposers = (/** @type {Set<Composer>} */ composers) => {
 	/** Class name of the first composer. */
-	let initialClassName = ''
+	let latestClassName = ''
 
 	/** @type {string[]} Combined class names for all composers. */
 	const combinedClassNames = []
@@ -313,7 +313,7 @@ const getPreparedDataFromComposers = (/** @type {Set<Composer>} */ composers) =>
 	const combinedUndefinedVariants = []
 
 	for (const [className, , , , prefilledVariants, undefinedVariants] of composers) {
-		if (initialClassName === '') initialClassName = className
+		latestClassName = className
 
 		combinedClassNames.push(className)
 
@@ -327,7 +327,7 @@ const getPreparedDataFromComposers = (/** @type {Set<Composer>} */ composers) =>
 
 	/** @type {[string, string[], PrefilledVariants, Set<UndefinedVariants>]} */
 	const preparedData = [
-		initialClassName,
+		latestClassName,
 		combinedClassNames,
 		combinedPrefilledVariants,
 		new Set(combinedUndefinedVariants)
--- a/packages/core/tests/component-composition.js
+++ b/packages/core/tests/component-composition.js
@@ -16,9 +16,9 @@ describe('Composition', () => {
 		const bold = css({ fontWeight: 'bold' })
 		const title = css(red, size14, bold, { fontFamily: 'monospace' })
 
-		expect(title.className).toBe('c-gmqXFB')
+		expect(title.className).toBe('c-kngyIZ')
 		expect(toString()).toBe('')
-		expect(String(title)).toBe('c-gmqXFB')
+		expect(String(title)).toBe('c-kngyIZ')
 		expect(toString()).toBe(
 			`--stitches{--:2 c-gmqXFB c-hzkWus c-cQFdVt c-kngyIZ}@media{` +
 				`.c-gmqXFB{color:red}` +
--- a/packages/core/tests/component-styled-composition.js
+++ b/packages/core/tests/component-styled-composition.js
@@ -1,0 +1,33 @@
+import { createCss } from '../src/index.js'
+
+describe('Composition', () => {
+	test('Renders an empty component', () => {
+		const { css, toString } = createCss()
+		const generic = css()
+
+		expect(generic().props).toEqual({ className: 'PJLV' })
+		expect(toString()).toBe('')
+	})
+
+	test('Renders composed styles with their own classes', () => {
+		const { css, toString } = createCss()
+		const box = css({ borderSizing: 'border-box' })
+		const flex = css(box, { display: 'flex' })
+		const row = css(flex, { flexDirection: 'row' })
+		expect(box.className).toBe('c-hkrHCb')
+		expect(flex.className).toBe('c-dhzjXW')
+		expect(row.className).toBe('c-ejCoEP')
+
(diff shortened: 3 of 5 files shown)


LucasUnplugged:
CodeSandbox showing the expected result: https://codesandbox.io/s/stitches-ci-cra-forked-5vkri
Versus CodeSandbox showing the issue: https://codesandbox.io/s/staging-field-4cs36

peduarte:
Thanks for this! I tested the sandbox and looks good to me. I think it would be super useful to create a dedicated test for this use case though. I don't think we have a test specifically for this. 

About the code, I'll leave that for @jonathantneal

LucasUnplugged:
> I think it would be super useful to create a dedicated test for this use case though. I don't think we have a test specifically for this. 

Yes, great point. I'm sure I can come up with some tests pretty easily.

peduarte:
Nice one. Thanks for that. We'll review shortly

jonathantneal:
Hey @LucasUnplugged, one issue with this change is that it moves the problem to the other side of composition.

See https://codesandbox.io/s/stitches-ci-cra-forked-y0u3u?file=/src/App.tsx

Now, the last class in a composition may return some unintended matches.

LucasUnplugged:
> Hey @LucasUnplugged, one issue with this change is that it moves the problem to the other side of composition.
> 
> See https://codesandbox.io/s/stitches-ci-cra-forked-y0u3u?file=/src/App.tsx
> 
> Now, the last class in a composition may return some unintended matches.

@jonathantneal I think that's a bit of a separate issue, and maybe not even an issue (sort of). What I mean by that is:
1. The issue happens with the current version, but instead of a sibling exhibiting the issue, a parent (or multiple) can, [as seen here](https://codesandbox.io/s/little-forest-hx1b3?file=/src/App.tsx).
2. As seen in both that sandbox, and in [this one (forked from yours)](https://codesandbox.io/s/stitches-ci-cra-forked-bzd97?file=/src/App.tsx), the issue is a factor of how the hashes are generated, I believe: same styles === same hash.
3. I say _"kind of"_ not an issue because of the fact that it's probably an unavoidable fact of the purely hash-based class generation, but having identical styles on two nearby components is probably pretty rare.

`<ShamelessSelfPromotion>`
That said, I think that [what I'm proposing here](https://github.com/modulz/stitches/discussions/677) would actually fix this issue, by creating classes that are a combination of a custom string + the generated hash; so you could have `c-UnderlinedBlue-hrMtea` and `c-UnderlinedPink-hrMtea`, and the hashes being the same won't cause a conflict :)
`</ShamelessSelfPromotion>`

peduarte:
@lucasmotta thanks for that! 🙏 

Jonathan and I had a long chat about this yesterday. The conclusion is that whichever way we go, the problem will still exist. And it depends on how users are creating + targeting their components (composition, descendant selector, etc)

> the issue is a factor of how the hashes are generated, I believe: same styles === same hash.

You're right. 

> I think that what I'm proposing here would actually fix this issue, by creating classes that are a combination of a custom string + the generated hash

Crating the hash from something more than just the style object would likely solve this issue. However, as I [explained in this discussion](https://github.com/modulz/stitches/discussions/677#discussioncomment-1086608), your current proposal isn't going to work as a third argument. 

I'm also concerned about the DX. As a user, I wouldn't want to have to give every element a friendly class name, especially when I am already naming the const. Feels like doing the same thing twice. This is problematic since it means the way to create more unique hashes would be to rely on babel, and this is something we don't want Stitches _to have to depend on_.

---

I think our options are:
- Leave it as is (issue exists in one end)
- Accept this PR (issue exists in the opposite end)
- Stop stacking classes (issue exists but since classes are not being stacked, it's a lot less likely to happen)

Thoughts?

LucasUnplugged:
> Crating the hash from something more than just the style object would likely solve this issue. However, as I [explained in this discussion](https://github.com/modulz/stitches/discussions/677#discussioncomment-1086608), your current proposal isn't going to work as a third argument. 

Right, in vanilla JS I guess that's true; I didn't think about that. In TS, it throws an error if you try to repeat element params.

That said, that's easily fixable by targeting the last param, and checking if it's a string. I think that would cover all TS and JS use-cases.

> I'm also concerned about the DX. As a user, I wouldn't want to have to give every element a friendly class name, especially when I am already naming the const. Feels like doing the same thing twice. This is problematic since it means the way to create more unique hashes would be to rely on babel, and this is something we don't want Stitches _to have to depend on_.

Yeah, I completely understand this sentiment. I think that a solution like mine doesn't preclude that, if we keep it a "hidden feature" and promote the babel plugin as a solution.

The reason why I think we may want a component of this solution to be in Stitches itself, is because we can then make sure these custom classes work exactly as classes currently do in Stitches, including stacking, and having the styles associated with the styles. Anyhow, I'll continue that discussion there.

> I think our options are:
> 
> - Leave it as is (issue exists in one end)
> 
> - Accept this PR (issue exists in the opposite end)
> 
> - Stop stacking classes (issue exists but since classes are not being stacked, it's a lot less likely to happen)

I'm still in favour of No.2, because No.1 has more than just this secondary issue (hash-generation): it also has the primary issue of breaking use of composed components inside nested style objects. I think that issue can be much more pervasive, in codebases that use composition, than two nearby components having identical styles.

Remember: the hash issue is only likely an issue for siblings and children of siblings, because of encapsulation. I don't see a real need to use components as selectors with global styles: just put those styles in those components — and if you want to avoid repetition, create a separate object for such styles and spread them on the component definition.

So saying all that, I think No.3 is probably unnecessary. I think the reason the hash issue hasn't come up until now is because that's a much more rare problem than the composition one, which could happen a lot in some codebases.

(Sorry for the long comment 😅)

LucasUnplugged:
BTW, if this gets approved, should I update the version numbers in my PR, or is that a whole separate process?

peduarte:
@LucasUnplugged Thank you 🙌 

To be honest, right now I am really struggling to believe that there's a difference between option 1 and 2 (first or last class composition). 

I _think_ I agree that with option 2 it's less likely to happen. 

> So saying all that, I think No.3 is probably unnecessary.

Why do you think that?

---

We need to wait to hear Jonathan's thoughts and decide whether we can merge this or not. 

To be transparent with you, our current priority is to get Stitches stable from an API perspective. This is more like a bug fix so it's not currently right at the top of our list. But I'll see if we can get an answer ASAP.

You can see our roadmap here: https://github.com/modulz/stitches/projects/4

LucasUnplugged:
Understood on the priority front.

For No.3, maybe I'm misunderstanding, but wouldn't that mean a whole revamp of how styles are associated with classes? If you have a stack of 3 classes on a component now, and want to reduce it to a single class, all those styles would have to get merged into that one class, right?

Maybe that's simpler to do than I'm imagining.

I'll try to create a sandbox to capture what I think is the difference between 1 and 2.

LucasUnplugged:
@peduarte Okay, I've made two sandboxes to try to highlight the differences:

1. [This sandbox](https://codesandbox.io/s/nervous-surf-6ir27?file=/src/App.tsx) uses the existing version of Stitches. You can see that the footer is not fine in any of the 3 blocks, because it should never have a purple background. The footer incorrectly inherits that style from both the cases where composition is at work, and cases where identical styles === identical hashes.

2. [This other sandbox](https://codesandbox.io/s/stitches-ci-cra-forked-4lxqx?file=/src/App.tsx) uses my version from this PR. The footer is correct in any case where only composition is at work. The hash issue is still present in the second block, but it can be addressed via a variant, as seen in block 3 — and the variant code wasn't even repeated in that solution.

So in summary, I think conflicts due to same hash are much easier to avoid than conflicts due to inheritance, in codebases that make heavy use of (primitive) composition. I've already seen this in the wild. Passing the exact same styles to two nearby components seems rare, and likely fixable via variants instead.

That's my 2 cents, anyway :)

jonathantneal:
In the current canary, we are using a `:where()` selector to target all of the elements in a composition without increasing specificity. It may resolve this issue, without choosing which side of the composition to use.

LucasUnplugged:
> In the current canary, we are using a `:where()` selector to target all of the elements in a composition without increasing specificity. It may resolve this issue, without choosing which side of the composition to use.

@jonathantneal Oh, very interesting! I tried a recent sandbox with my example above, and it does seem to fix the issue (🎉); but I'm having trouble with nesting of styled components. Did the syntax for that changes? It seems to be working, by Typescript is screaming at me: https://codesandbox.io/s/objective-sound-57d99?file=/src/App.tsx

So before, I used to do something like this:
```TS
const Main = styled(Box, {
  background: '$gray500',
  [`& ${Header}`]: {
    background: '$purple600',
    color: 'white',
  },
});
```
(Where `Header` is a Stitches-styled component)

peduarte:
Removing this from the V1 roadmap. We'll have to come back to this after launch, with more time, to fully understand the issue, side effects, proposed solutions, etc.

LucasUnplugged:
> Removing this from the V1 roadmap. We'll have to come back to this after launch, with more time, to fully understand the issue, side effects, proposed solutions, etc.

Actually, v1 seems to have fixed this issue, @peduarte! And the type issue I last mentioned above also seems to have been fixed, since the earlier canary versions 🎉 [Sample sandbox](https://codesandbox.io/s/stitches-ci-cra-forked-zwdqr?file=/src/App.tsx) showing it working just fine.

As such, I'm closing this PR.

peduarte:
Glad to hear :smi

Issue #713, opened 2021-08-10
TS:1.0 - createTheme return type does not include ThemeTokens when theme class is used


jake-nz:
Without a theme class it's fine: `createTheme(theme)` 

When you try to define the theme class by passing it as the first argument: `createTheme('dark-theme', theme)` the return type is just `{ className: string; selector: string; } & "dark-theme"`

## To Reproduce
```js
const { createTheme } = createStitches({
  theme: { colors: { ...gray } }
})

export const dark = createTheme('dark-theme', {
  colors: { ...grayDark }
})

console.log(dark.colors.gray1)
```
> Property 'colors' does not exist on type '{ className: string; selector: string; } & "dark-theme"'

peduarte:
Hi! This has been released to Stitches v1.1.0

```
yarn add @stitches/react@latest
# or
npm install @stitches/react@latest
```

PR #714, opened 2021-08-10
fix: return type for createTheme

jake-nz:
fixes: modulz/stitches#713

Check if arg0 is a string (theme class name), if so, include tokens from arg1, otherwise, include tokens from arg0


peduarte:
Nice one. Thanks for that. @jonathantneal can you review?

jonathantneal:
Thank you, @jake-nz !

PR #721, opened 2021-08-11
[Feature] Dynamic variants

ch99q:
This introduces a new set of features in stitches called dynamic variants.

The purpose of this is to allow dynamic props to apply CSS values at static, server, and client-side.

Example of dynamic spacer component.
```jsx
export const Spacer = styled("div", {
  $$spaceHeight: 0,
  $$spaceWidth: 0,
  
  minHeight: "$$spaceHeight",
  minWidth: "$$spaceWidth",

  variants: {
    size: (size: Stitches.ScaleValue<'space'>) => {
      $$spaceHeight: size,
      $$spaceWidth: size,
    },
    height: (value: Stitches.ScaleValue<'space'>) => {
      $$spaceHeight: value,
    },
    width: (value: Stitches.ScaleValue<'space'>) => {
      $$spaceWidth: value,
    }
  }
})
```


packages/core/types/stitches.d.ts
```diff
@@ -162,7 +168,7 @@ export default interface Stitches<
 			}
 		): StyledComponent.CssComponent<
 			StyledComponent.StyledComponentType<Composers>,
-			StyledComponent.StyledComponentProps<Composers>,
+			StyledComponent.StyledComponentProps<Composers, Media> & { css?: CSS },
```
ch99q, line 171: Should remove the css prop, due to new improved way.
```suggestion
			StyledComponent.StyledComponentProps<Composers, Media>,
```

jonathantneal:
It looks like there are still some conflicts to work out. Starting with a clean copy of `stitches.d.ts` and restoring your modifications (if any) from there might be easier. Let me know if you’d like some additional help. 😄

ch99q:
Okay so I finally got the typings to work properly, now I just have to ensure the tests run perfectly.

Well, I have a discussion with myself about the arguments of a function, but I realize this also includes utility functions.
So lets say we have a function e.g.
```ts
variants: {
    size: (value: Stitches.ScaleValue<'space'>) => ({
        width: value,
        height: value
   })
}
```
This should result in a component with the prop `size="$small"` a single argument. But what if a function has more than a single argument like.
```ts
variants: {
    translate: (x: number, y: number) => ({
        translate: `${x}px, ${y}px`
   })
}
```
Currently, the typings only support the first argument as prop value (this also applies to utils), but should we look into allowing for multiple arguments through `translate={[100, 200]}`?

It will not be a part of this pull request, but a general question about the future of utility functions and dynamic variants.

fmal:
@ch99q this is going to be great addition to stitches, i can think of so many cases where this functionality would be useful, brilliant work :) @peduarte pls bring it into the lib when the time is right 😃

peduarte:
It is indeed a good feature. We're having lots of internal discussions about the future of stitches.

Right now, this isn't a priority. Stitches goal has always been to have as minimal runtime as possible. This increased the potential amount of runtime so we have to think about how it fits in the roadmap, how it affects performance, etc. 

But it is being considered!

manspaniel:
I really hope this or something like it makes it into the lib soon! It feels like the final piece of the puzzle.

Something I've found while working on a few sites with Stitches is that it's quite cumbersome to fudge this behaviour by creating a wrapper functional component and using a `css` prop in that wrapper — forwarding refs, merging the css prop, getting prop types for tokens etc. Because of this, use of the `css` prop becomes a necessary evil.

Understand the performance concern though, especially after coming from `theme-ui` where there are some serious performance issues on complex pages. Performance wise, is this much different from using the `css` prop? Is the main concern just that this encourages too much dynamic behaviour? I'd argue that the _lack_ of this feature encourages overuse of the `css` prop in general — if I have to define some one-off flex settings for a Box via the `css` prop, I may as well add my padding/border/font/color styles there too. At least with this proposed feature, we can be more intentional about which dynamic values a component can receive, and I assume it could be optimised more easily by the library as well, since you could potentially regenerate styles + memoize on a per-variant basis.

Congrats on v1! 🎉

darklight9811:
Really looking forward into this, since I'm building my entire design system with stitches, having dynamic column sizes is really a hassle without anything like dynamic variants.

peduarte:
Noted 🙏 and thank you!

lukemorales:
I'm on board with the @fmal @manspaniel and @darklight9811 arguments! Stitches have already proven to be one of the best, if not THE best, ways to create design systems.
And while I understand that runtime is important and it's one of the greatest features of Stitches, I believe that the smashing majority of users fall in love with Stitches because of the awesome DX your team was able to achieve, and dynamic variants vastly improve the DX for a lot of use cases that, without it today, we are limited to what we can achieve (I think a great example is the Stack component for Radix UI, I've seen that in order to achieve a fully-featured component, the team had to create every possible variant for flex-direction values, justify-content values, align-items values, etc.). And this limitation breaks a lot of that great DX Stitches has because we can only go so far before hitting a wall, and dynamic variants definitely would have improved the DX on that Stack component or for the use case that @darklight9811 just gave us.

Just adding another argument in favor of this feature even if we compromise a little on the runtime

darklight9811:
@lukemorales yes, but I kinda understand if they decide not do it, but making my own design system, I made a util that would build everything for me, so it wasn't a deal breaker after all.

lukemorales:
@darklight9811 yeah, I would understand as well if they decide not to do it, but the reason I believe the smashing majority of users wanna use Stitches is because of the awesome DX, the awesome runtime is a plus, and dynamic variants definitely improves DX for components like this: (Flex Component in Radix Design System)[https://github.com/radix-ui/design-system/blob/master/components/Flex.tsx] to be written like this:

```js
export const Flex = styled('div', {
  boxSizing: 'border-box',
  display: 'flex',

  variants: {
    direction: (value: TypeForCSSPropertyValues<'flexDirection'>) => ({
      flexDirection: value,
    }),
    align: (value: TypeForCSSPropertyValues<'alignItems'>) => ({
      alignItems: value,
    }),
    justify: (value: TypeForCSSPropertyValues<'justifyContent'>) => ({
      justifyContent: value,
    }),
    wrap: (value: TypeForCSSPropertyValues<'flexWrap'>) => ({
      flexWrap: value,
    }),
    gap: (value: Stitches.ScaleValue<'space'>) => ({
      gap: value,
    })
  },
  defaultVariants: {
    direction: 'row',
    align: 'stretch',
    justify: 'start',
    wrap: 'noWrap',
  }
});
```

darklight9811:
@lukemorales one thing that I didn't see how they are going handle is the type of the generated variant, since they do this automatically because of the predefined scope, how they would generate that for unknown values?

lukemorales:
> @lukemorales one thing that I didn't see how they are going handle is the type of the generated variant, since they do this automatically because of the predefined scope, how they would generate that for unknown values? And most dynamic variants would still need a constraint for that (both type and callback), would that be handled manually by the programmer? If so, that would kinda sacrifice a part of DX to get another part.

Maybe I'm not understanding what you're trying to describe, but I don't see how this would be different from the config utils Stitches already has?

darklight9811:
Never mind, just noticed they could use the argument type to type the prop. For example, if you defined a column dynamic variant, just use the parameter for the type `(value: 1 | 2 | 3 | 4)` and constraint it with a clamp method inside it.

ch99q:
I would like to say theres possibilities to improve the runtime performance, with features line predefined arguments (Making it generate styles for an array og argument possibilities).

Also the runtime performance is only affected when changing arguments in runtime.

When hydrating the application in SSG or SSG it will generate the styles at that point.

pleunv:
Just started getting my feet wet with Stitches, coming from a rather large emotion + styled-system codebase. One of the first things I was also looking for was a way to map variants to a whole scale of theme tokens or valid set of css properties without having to manually map every individual value so that we could easily port over our primitives. As far as I can tell there is currently no way to achieve this, which would make this a rather cumbersome migration (and possibly also maintenance) effort.
I think this is exactly what this proposal is suggesting so definitely a thumbs of from me as well.

joe-bell:
@pleunv love that `createScaleVariant` concept – that's exactly the kind of API I'm after (see #816) 💪

peduarte:
@pleunv replied here https://github.com/modulz/stitches/discussions/816#discussioncomment-1423958

pstoica:
@emadabdulrahim This is the same pattern I came to before realizing there was a request for dynamic variants. I feel like having this in the docs as a recipe would be great.

While shorthand could be nice, I don't personally feel like I'm missing anything from stitches using static objects as long as I can use my token definitions as types.

LucasUnplugged:
> Something I've found while working on a few sites with Stitches is that it's quite cumbersome to fudge this behaviour by creating a wrapper functional component and using a `css` prop in that wrapper — forwarding refs, merging the css prop, getting prop types for tokens etc. Because of this, use of the `css` prop becomes a necessary evil.
> 
> Understand the performance concern though, especially after coming from `theme-ui` where there are some serious performance issues on complex pages. Performance wise, is this much different from using the `css` prop? Is the main concern just that this encourages too much dynamic behaviour? I'd argue that the _lack_ of this feature encourages overuse of the `css` prop in general — if I have to define some one-off flex settings for a Box via the `css` prop, I may as well add my padding/border/font/color styles there too. 

So I was struggling with this kind of usage a bit at first; but since I'm creating a whole design system, with extensive theme tokens (including over 30 size/spacing tokens), I was ultimately able to use variants to cover most cases, without requiring this kind of runtime-centric solution.

For instance, here is a bit of code I'm producing, via Stitches-powered primitives (NOTE: my variants are based on `styled-system`):
```JSX
      <Row alignItems="center">
        <Box h="7" w="6" bg="primary9" radiusLeft="4" mr="px" />
        <Box h="7" w="6" bg="secondary9" radiusRight="4" mr="3" />
        <Heading bold flat h="7" mb="1" color="secondary9">
          {title}
        </Heading>
      </Row>
```
Which produces this output:
Screen Shot 2021-12-11 at 3 19 44 PM

I've created some helper functions to help me generate sequences of variants, without writing them all out, but having them fully typed — e.g.:
```TS
    // Produces a sequence of 12 numbered fontSize variants, starting at 1
    fontSize: generateVariantSequence<Size>(12, (n: number) => ({
      fontSize: `$${n}`,
    })),
```

You can even extend it in some ways that really simplify some complex cases. For instance, I created a `Grid` primitive that's insanely powerful, yet really simple (and fully typed):
```JSX
      <Grid columnFit="11" p="5" radius="2">
        {keys.map((key: number) => (
          <GridItem
            key={key}
            bg={`primary${key as ColorNumberKey}`}
            color={`textPrimary${key as ColorNumberKey}`}
            p="4"
          >
            {inner}
          </GridItem>
        ))}
        {keys.map((key: number) => (
          <GridItem
            key={key}
            bg={`secondary${key as ColorNumberKey}`}
            color={`textSecondary${key as ColorNumberKey}`}
            p="4"
          >
            {inner}
          </GridItem>
        ))}
      </Grid>
```

Which produces this output:
Screen Shot 2021-12-11 at 3 39 34 PM

The heavy lifting `columnFill` variant is coded as:
```TS
    columnFill: generateVariantSequence<FillVariant>(
      15,
      (n: number) => ({ gridTemplateColumns: `repeat(auto-fill, minmax($sizes$${n}, 1fr))` }),
      4
    ),
```

If you like the DX of something like Chakra, but want the performance and customizability of Stitches, I think this is a great solution.

---

Between something like what I'm doing, and the creation of your own primitives — which can map additional dynamic props to `css={{...}}`, I personally think it's best to actually discourage this feature (no offense to @ch99q's excellent and hard work).

One might argue and say, "well, you don't need to use it, even if they put it in", which is true, but...
1. That's more JS that will get bundled and shipped, which many folks would have no use for.
2. It will encourage using Stitches in a way that, to my understanding, was against to whole original premise. At that point, it becomes harder to argue your team should switch to Stitches for performance reasons, when there's a usage of it that will degrade performance as well (you'd have to introduce conventions against that... it's a whole mess).
3. That usage will feel more natural to anyone coming from more dynamic, and less performant, styling libs, so Stitches would be contributing to the issue, instead of promoting better alternatives.

Anyhow, that's my 2 cents.

fmal:
@LucasUnplugged really nice examples, would you mind sharing a codesandbox with the implementation of `generateVariantSequence`, the `Grid` component etc? Would be a nice reference for anyone wanting to leverage the full power of variants 😃

LucasUnplugged:
> @LucasUnplugged really nice examples, would you mind sharing a codesandbox with the implementation of `generateVariantSequence`, the `Grid` component etc? Would be a nice reference for anyone wanting to leverage the full power of variants 😃 

Yeah, I'd be happy to! I'll post here once it's ready.

LucasUnplugged:
As promised, here's a sandbox showcasing some of what I mentioned above; it's not a super complex example, and the typing isn't perfect (I spent quite a while on this in my design system), but it should do the trick:

https://codesandbox.io/s/stitches-variants-showcase-wtr3i?file=/src/App.tsx

nihgwu:
I followed @emadabdulrahim 's idea and created [this demo](https://codesandbox.io/s/gifted-fast-fyt39?file=/src/App.tsx), though I'm not sure it's a good idea as it will generate a huge style for each component, as I have to use `initial` to prevent nested inherits which I think is an issue in @emadabdulrahim 's demo

elsangedy:
```ts
const { config } = createStitches({
  theme: {
    colors: {
      primary: 'red',
    },
    space: {
      1: '',
      2: '',
      3: '',
    },
  },
});

type CSS = Stitches.CSS<typeof config>;

export const getVariant = <
  P extends keyof typeof config.theme,
  T extends keyof typeof config.theme[P],
  R extends Record<T, CSS>,
>(
  prop: P,
  map: (tokenValue: `$${T}`) => CSS,
): R => {
  const values = Object.keys(config.theme[prop]) as T[];
  return values.reduce<R>(
    (acc, tokenValue) => ({ ...acc, [tokenValue]: map(`$${tokenValue}`) }),
    {} as R,
  );
};

const Flex = styled('div', {
  boxSizing: 'border-box',
  display: 'flex',
  variants: {
    color: getVariant('colors', (tokenValue) => ({
      color: tokenValue,
    })),
    gap: getVariant('space', (tokenValue) => ({
      gap: tokenValue,
    })),
  },
});

// <Flex color="primary" gap={2} />
```

rudeayelo:
@elsangedy seems like TypeScript has some issues with your example: https://codesandbox.io/s/priceless-shape-6i7ku?file=/src/App.tsx

❤️ the solution by the way

f1lander:
> ```ts
> const { config } = createStitches({
>   theme: {
>     colors: {
>       primary: 'red',
>     },
>     space: {
>       1: '',
>       2: '',
>       3: '',
>     },
>   },
> });
> 
> type CSS = Stitches.CSS<typeof config>;
> 
> export const getVariant = <
>   P extends keyof typeof config.theme,
>   T extends keyof typeof config.theme[P],
>   R extends Record<T, CSS>,
> >(
>   prop: P,
>   map: (tokenValue: `$${T}`) => CSS,
> ): R => {
>   const values = Object.keys(config.theme[prop]) as T[];
>   return values.reduce<R>(
>     (acc, tokenValue) => ({ ...acc, [tokenValue]: map(`$${tokenValue}`) }),
>     {} as R,
>   );
> };
> 
> const Flex = styled('div', {
>   boxSizing: 'border-box',
>   display: 'flex',
>   variants: {
>     color: getVariant('colors', (tokenValue) => ({
>       color: tokenValue,
>     })),
>     gap: getVariant('space', (tokenValue) => ({
>       gap: tokenValue,
>     })),
>   },
> });
> 
> // <Flex color="primary" gap={2} />
> ```

This was mt approach

```
import "./styles.css";
import type * as Stitches from "@stitches/react";
import { styled, theme, CSS, css } from "./stitches.config";

const { colors } = theme;

type TColors = {
  [K in keyof typeof colors]: { $$color: string };
};

const ColorsVariants = css({
  variants: {
    color: {
      inherit: {
        $$color: "currentColor"
      },
      ...Object.keys(colors).reduce(
        (prev, curr) => ({ ...prev, [curr]: { $$color: "$colors$" + curr } }),
        {} as TColors
      )
    }
  }
});

const SvgIcon = styled(
  "svg",
  {
    $$size: "1em",
    $$color: "$colors$black",
    lineHeight: "1em",
    verticalAlign: "middle",
    width: "$$size",
    height: "$$size",
    // here we actually select the path data with css
    "& path": {
      stroke: "$$color"
    },
    variants: {
      size: {
        xs: {
          $$size: "10px"
        },
        xl: {
          $$size: "$sizes$xl"
        }
      }
    }
  },
  ColorsVariants
);

type TSvgIcon = {
  css?: CSS;
};

const MyCoolIcon = (
  props: TSvgIcon & Stitches.VariantProps<typeof SvgIcon>
) => (
  <SvgIcon
    viewBox="0 0 10 10"
    fill="none"
    xmlns="http://www.w3.org/2000/svg"
    {...props}
  >
    <path
      d="M8.74997 1.25L1.25 8.74997M1.25 8.74997L8.75 8.75M1.25 8.74997L1.25003 1.25003"
      stroke="currentColor"
      strokeLinecap="round"
      strokeLinejoin="round"
    />
  </SvgIcon>
);

export default function App() {
  return (
    <div className="App">
      <h1>Hello CodeSandbox</h1>
      <h2>Start editing to see some magic happen!</h2>
      <MyCoolIcon size="xs" color={{ "@xs": "inherit", "@sm": "red" }} />
    </div>
  );
}
```

https://codesandbox.io/s/thirsty-satoshi-bbggf?file=/src/App.tsx:0-1652

I will try yours because seems more clever in matter to type the key of the locale token

hadihallak:
in case anyone is interested in this, I did a little experiment to showcase how to handle highly dynamic variants using css custom properties and inline styles. you can check it out [here](https://m4gwwc.csb.app/)

PR #752, opened 2021-09-01
Fix: Allow string array type for global at import rule

Zwooosh:
Since `v.1.0.0` the allowed type of `string[]` for `@import` rules was removed from `globalCss`.
The actual functionality remained unchanged, because multiple `@import` rules still work. 

This PR re-adds support for `string[]` types inside `globalCss`.


jonathantneal:
Wonderful, thank you, @Zwooosh !

Issue #758, opened 2021-09-04
Error when adding duplicate identifiers in CSS objects


kiastorm:
# Bug report

## Describe the bug
Doing a 100vh fix based on this article, https://css-tricks.com/css-fix-for-100vh-in-mobile-webkit/ , and trying to put two rules , as instructed, I get an error.

```
An object literal cannot have multiple properties with the same name in strict mode.ts(1117)
Duplicate identifier 'minHeight'.ts(2300)
```

## To Reproduce

```
stitchesConfig.globalCss({
  body: {
    margin: 0,
    minHeight: '100vh',
    minHeight: '-webkit-fill-available',
  },
  html: {
    height: '-webkit-fill-available',
  },
});
```

## Expected behaviour
I'm not sure why this is required, but in this scenario I'd expect to be able to pass an array in order to set multiple rules for a single CSS property. Similar to: `'@font-face': [{}, {}, {}]`, I'd expect to be able to do :
```
body: {
  minHeight: ['100vh', '-webkit-fill-available']
}
```

## Screenshots

![image](https://user-images.githubusercontent.com/63059016/132092539-33903e00-8f19-413e-a3ec-f2079cfda4ec.png)

peduarte:
Thanks for raising this issue. Unfortunately, that's a limitation of the JS object syntax. 

With the `styled/css` APIs, we allow `n` Style Objects arguments. That way, you can pass them as separate args and Stitches will merge them for you:
```jsx
const Button = styled('button', { minHeight: '100vh' }, { minHeight: '-webkit-fill-available' })
```

But the `globalCss` function only accepts one. I'm thinking whether it's a good idea to allow `n` as well. 

Would that be a good solution for you?

Meanwhile, if you need this to work today, you can do:
```jsx
globalCss({
  body: {
    // hack into the object string as raw CSS
    minHeight: '100vh; min-height: -webkit-fill-available;' 
  }
})
```

Note: styled-components allow infinite args for their global css API https://codesandbox.io/s/funny-frog-txtcm?file=/src/App.tsx

/cc @jonathantneal

kiastorm:
> ```js
> const Button = styled('button', { minHeight: '100vh' }, { minHeight: '-webkit-fill-available' })
> ```
>
> ```js
> globalCss({
>   body: {
>     // hack into the object string as raw CSS
>     minHeight: '100vh; min-height: -webkit-fill-available;' 
>   }
> })
> ```

These make sense and work for me, thanks!

[...]

PR #762, opened 2021-09-06
Support n composition in `globalCss` function

jonathantneal:
This PR changes the `globalCss` function to support multiple arguments of global styles. It also updates the types to support this behavior.

Resolves https://github.com/modulz/stitches/issues/758


peduarte (approved):
Thanks!

PR #793, opened 2021-09-17
Fix TS issue with globalCss

jonathantneal:
This resolves a TypeScript issue where multiple styles passed into `globalCss` throw a typing error.

Resolves https://github.com/modulz/stitches/issues/794


peduarte (approved):
Cant test because CSB is down, but let's go.

PR #822, opened 2021-10-01
Add distributive condition to `Assign` util

### LINKED ISSUE #821: Property `collapsible` does not exist when composing radix-ui Accordion with Stitches
# Bug report

## Describe the bug

Property `collapsible` on the `Root` component of radix-ui's Accordion does not exist when composing it with Stitches

## To Reproduce

https://codesandbox.io/s/inspiring-sea-73jfp?file=/src/App.tsx

## Expected behavior

Property `collapsible` should exist when composing radix-ui Accordion with Stitches

jjenzz:
Fixes #821 by adding support for components with union prop types.

https://www.typescriptlang.org/docs/handbook/advanced-types.html#distributive-conditional-types

### DIFF
--- a/packages/react/tests/types.test.tsx
+++ b/packages/react/tests/types.test.tsx
@@ -166,3 +166,19 @@ void function Test() {
 		}} />
 	)
 }
+
+/* -------------------------------------------------------------------------------------------------
+ * Issue #821
+ * -----------------------------------------------------------------------------------------------*/
+
+type UnionProps = { type: 'single', collapsible: boolean; } | { type: 'multiple' };
+const UnionComponent: React.FC<UnionProps> = () => null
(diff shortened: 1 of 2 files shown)


packages/react/tests/types.test.tsx
```diff
@@ -2,7 +2,7 @@
 import * as Stitches from '../types/index'
 import { createStitches } from '../types/index'
 
-const { css, global, keyframes, styled, theme } = createStitches({
+const { css, globalCss, keyframes, styled, theme } = createStitches({
```
jjenzz, line 5: The test file was erroring because `global` doesn't exist so fixed this while I was here too.

jonathantneal (approved):

jonathantneal:
Thanks, @jjenzz!

Issue #826, opened 2021-10-04
css prop type doesn't include CssComponent


hudecsamuel:
# Bug report

## Describe the bug

Using `CssComponent` (return type of `css()`)  in styled component `css` prop produces an error
```tsx
import { styled, css } from "./stitches.config";

const exampleCss = css({
  padding: "10px"
});

const Box = styled("div", {});

export default function App() {
  return (
    <Box css={exampleCss}>
      Example
    </Box>
  );
}
```

The error: 
```
No overload matches this call.
  Overload 1 of 3, '(props: Assign<Pick<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, "color" | "translate" | ... 252 more ... | "onTransitionEndCapture"> & { ...; }, TransformProps<...> & { ...; }>): ReactElement<...> | null', gave the following error.
    Type 'CssComponent<never, {}, { bp1: "@media (min-width: 520px)"; bp2: "@media (min-width: 900px)"; }, CSS<{ bp1: "@media (min-width: 520px)"; bp2: "@media (min-width: 900px)"; }, { colors: { hiContrast: ...; ... 12 more ...; purple600: ...; }; space: { ...; }; sizes: { ...; }; fontSizes: { ...; }; fonts: { ...; }; }, Def...' is not assignable to type 'CSS<{ bp1: "@media (min-width: 520px)"; bp2: "@media (min-width: 900px)"; }, { colors: { hiContrast: ...; loContrast: ...; gray100: ...; gray200: ...; gray300: ...; gray400: ...; gray500: ...; gray600: ...; ... 5 more ...; purple600: ...; }; space: { ...; }; sizes: { ...; }; fontSizes: { ...; }; fonts: { ...; }; }, ...'.
[...]

PR #827, opened 2021-10-04
fix: added ComponentProps to css prop type

hudecsamuel:
Added `CssComponent<Type, Props, Media, CSS>` to `StyledComponent` `css` prop type

fixes https://github.com/modulz/stitches/issues/826


peduarte:
This PR fixes the type error, but this isnt a feature - in other words, it will never work (unless implemented)

More info in the issue.

PR #829, opened 2021-10-05
Update `PropertyValue` and `ScaleValue` types to support config

### LINKED ISSUE #813: Ability to type consts as Property or Scale values
We need a way to type consts as Property or Scale values outside of utils.

In an ideal world, I'd like us to reuse the existing APIs, `Stitches.PropertyValue` and `Stitches.ScaleValue`.

Use case:

Without config
```tsx
const ColorProps: Stitches.PropertyValue<'color'>
const color: ColorProps = 'RebeccaPurple'
```

With config
```tsx
// import { config } from 'stitches.config'
const ColorProps: Stitches.PropertyValue<'color', typeof config>
const color: ColorProps = '$pruple5'
```

Ideally the current usage within utils will remain the same. 

Happy to chat about alternate API's.

peduarte:
Related #813 

/cc @raunofreiberg

### DIFF
--- a/packages/core/types/index.d.ts
+++ b/packages/core/types/index.d.ts
@@ -33,10 +33,29 @@ export type CSS<
 export type ComponentProps<Component> = Component extends ((...args: any[]) => any) ? Parameters<Component>[0] : never
 
 /** Returns a type that expects a value to be a kind of CSS property value. */
-export type PropertyValue<K extends keyof CSSUtil.CSSProperties> = { readonly [CSSUtil.$$PropertyValue]: K }
+export type PropertyValue<K extends keyof CSSUtil.CSSProperties, C = null> = (
+	C extends null 
+		? { readonly [CSSUtil.$$PropertyValue]: K } 
+	: C extends { [K: string]: any } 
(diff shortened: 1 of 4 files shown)


raunofreiberg:
Yoo amazing! 🙏

jonathantneal (approved):

fmal:
@peduarte so it means that in my `stitches.config.ts` i could now export:

```ts
export type CSS = Stitches.CSS<typeof config>;
export type PropertyValue<T extends keyof Stitches.CSSProperties> =
  Stitches.PropertyValue<T, typeof config>;
export type ScaleValue<T> = Stitches.ScaleValue<T, typeof config>;
```
to apply the config and then in my components use it like this:
```tsx
function Component({ width }: { width: PropertyValue<'width'> }) {
  ....
}
```
Cool 😃 In such situations i have used the `CSS['cssProp`]` type so far

peduarte:
Exactly. `CSS['cssProp']` is the way to go if you havent created your own Stitches. But if you have a config (tokens), now you can use the new types to also have tokens being suggested

Issue #628, opened 2021-06-14
Stitches 0.2 - insertionMethod alternative for dynamically created <iframe> and shadow DOM


cpakken:
In Stitches 0.2, the `insertionMethod` is removed because of the transition to CSSOM. With that change, I propose some alternative API's that would grant the ability to initialize the root dom node dynamically / lazily.

- Use stitches styled components within dynamically created iframes and shadow doms
- can isolate styles using separate `createCss()` for different iframes / shadow doms 

Taking a look at the source code for Stitches 0.2, It seems that ```createCss({root: DOCUMENT})``` :
is an undocumented api for applying a custom root. However, this does not apply when the root is lazily / dynamically created.

## Use Case

Building a chrome extension with UI overlay. A shadowdom/ iframe is dynamically created to separate css / dom implementation to avoid clashing. 

### Method 1: Similar to emotion's CacheProvider

```jsx
//Emotion API for inspiration
<CacheProvider value={createCache({container: RootNode})}>
```

```jsx
const { styled, RootProvider } = createCss({})

const Box = styled('div', {})

const IFrameExample = () => {
  const ref = useRef()

  return (
    <iframe ref={ref}>
      <RootProvider root={ref.current.contentWindow.document}>
        {/* Simplified for example, uses portals to insert components into body */}
        <Box>Example</Box>
      </RootProvider>
    </iframe>
  )
}
```

### Method 2: Dynamically call initRoot

```jsx
const { styled, initRoot } = createCss({})

const Box = styled('div', {})

const IFrameExample = () => {
  const ref = useRef()
  useEffect(() => initRoot(ref.current.contentWindow.document), [])

  return (
    <iframe ref={ref}>
      {/* Simplified for example, uses portals to insert components into body */}
      <Box>Example</Box>
    </iframe>
  )
}
```

CraigCav:
This seems related to #395.

If you could scope a style sheet to a custom root (a given element in the react tree), then you'd be able to solve scoping issue for iframes and SSR (one sheet per request) at the same time.

hennessyevan:
While I like being explicit as in these two api proposals, perhaps we could just detect if we're in a shadow DOM or iframe and apply styles to the root there, adding `:host`, injecting the style tag into the shadow root, etc.

LexSwed:
Just in case it can give some inspiration, I'm not proud of this (I hoped it's temporary), but basically I provide an "env variable" when building my extension with the id of the element to insert the styles into:
[...]

PR #836, opened 2021-10-06
Remove `root` from `config`

peduarte:


jlalmes:
The `root` config options is _required_ for `shadowRoot` support and should not have been removed! (cc. https://github.com/modulz/stitches/issues/628 & https://github.com/modulz/stitches/discussions/747)

The real bug is on the following line, invoked by the `createMemo()` function:
https://github.com/modulz/stitches/blob/0ebaf9f988871ac0d8d5f2b72f2a8042e0d1b56f/packages/core/src/utility/createMemo.js#L1

Will open a PR to fix now.

jlalmes:
Opened a PR here: https://github.com/modulz/stitches/pull/1004

Issue #756, opened 2021-09-03
DOMException: Failed to read the 'cssRules' property from 'CSSStyleSheet' (file system based access)


sgrtho:
# Bug report

## Describe the bug

Environments using local-style resources (think of `file://`) might fail initialization because in this context scripts lack permissions to access `.cssRules` of linked stylesheets.

The topic is covered here:

https://stackoverflow.com/questions/48753691/cannot-access-cssrules-from-local-css-file-in-chrome-64

## To Reproduce

Since this is primarily an *environment*-related (and probably rare) issue, I won't provide a working example. But I think it is obvious that the issue originates from...

https://github.com/modulz/stitches/blob/canary/packages/core/src/sheet.js LINE 36 and 38

A  code snippet to reproduce this, when accessed e.g. in Chrome via local file system access:

    <html>
      <body>
        <link href="./path-to/some.css" rel="stylesheet">
        <script src="./path-to/some-script-launching-stitches.js"></script>
      </body>
    </html>

## Concerns

While this might seem like a rare issue, I want to express some concerns about how this could affect special environments like e.g. hybrid apps on iOS relying on local access.

**UPDATE:** I can confirm that this issue affects iOS when using *WKWebView* with `loadFileURL`/`allowingReadAccessToURL` IF there's a local stylesheet linked.

## Expected behavior

Improved detection of `.cssRules` access rights.

## System information

- OS: macOS
- Chrome 
- Version of Stitches: 1.0.0

sgrtho:
Is there a specific reason you're going through every stylesheet to find a hydratable one?

jonathantneal:
Hey @sgrtho, we decided to use the CSSOM exclusively for hydrating stylesheets because it offered greater performance. However, `file:` & [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) restrictions may prevent us from reading CSS files, if they are created for Stitches.

The _easy_ solution would be to ignore all stylesheets with an href. However, we’ve always wanted Stitches to work if users were able to export their styles into a file. In other words, "inline stylesheets only" were never our end-game, which means the _easy_ solution would immediately break anyone exporting styles to files. However again, this would cause issues for anyone in `file:` or **CSP** environments. Another _easy_ solution would be not explicitly support `file:` or **CSP**. A less-performant solution would be to wrap each check in a try/catch block. However, we’d need to move all other logic out of the try/catch, lest we push it all into catchable memory.

[...]

PR #853, opened 2021-10-20
Improve check for cssRules access

megawebmaster:
Hi! :wave: 

I was hit by a bug with checking access for `cssRules` property on existing CSS style or link tags: https://github.com/modulz/stitches/issues/756 It was a real problem for me as I'm introducing Stitches to my project (it uses existing styles for other parts of the system) and we rely on Storybook (which renders components in an iframe) to get the UI elements right.

To avoid having problems I propose to improve how Stitches check for access to `cssRules` property on `CSSStyleSheet` objects to also include a try to access them and bail out if we catch an exception based on how this object should behave: https://drafts.csswg.org/cssom/#the-cssstylesheet-interface

I decided to keep earlier check as it should be easier and faster to process in most cases (as it is working in current code).

### DIFF
--- a/packages/core/src/sheet.js
+++ b/packages/core/src/sheet.js
@@ -5,6 +5,19 @@
 /** @type {RuleGroupNames} */
 const names = ['themed', 'global', 'styled', 'onevar', 'allvar', 'inline']
 
+const isSheetAccessible = (/** @type {CSSStyleSheet} */ sheet) => {
+	if (sheet.href && !sheet.href.startsWith(location.origin)) {
+		return false
+	}
+
+	try {
+		sheet.cssRules
+		return true
+	} catch (e) {
+		return false
+	}
+}
+
 export const createSheet = (/** @type {DocumentOrShadowRoot} */ root) => {
 	/** @type {SheetGroup} Object hosting the hydrated stylesheet. */
 	let groupSheet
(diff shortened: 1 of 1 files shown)


megawebmaster:
@peduarte - how is this going? I'm a bit worried, as I received a few production errors from Rollbar about "SecurityError: CSSStyleSheet.cssRules getter: Not allowed to access cross-origin stylesheet". 

If you still need some additional time I'll just use my branch for the time being and switch back when the issue is resolved :+1:

peduarte:
@hadihallak would you be able to review this if you have a chance? Thanks!

hadihallak:
@megawebmaster Thank you so much for looking into this and working on a fix 🙏 
will be reviewing it later today.

hadihallak:
Hey @megawebmaster 
the code looks great aside from some indentation changes but I would be very helpful if you could provide a minimal reproduction of the issue as I think I might be able to handle it elsewhere and having a reproduction will help me make sure it's tested in the codebase.
 
Thanks again for submitting the PR. Very appreciated.

megawebmaster:
Hey @hadihallak

I will try to prepare something for you.

hadihallak:
@megawebmaster Amazing! thank you so much 🙏

megawebmaster:
@hadihallak - here's a repo with the issue: https://github.com/megawebmaster/stitches-cssrules-reproduction

I added all descriptions how to run it there, let me know if you need any help.

Anyway: I distilled the issue thanks to that to invalid `<link>` tag that appears first inside a frame.

hadihallak:
@megawebmaster Thanks! This is going to be super helpful.
will be taking another look early next week just as soon as we sort out some of our high priority work around injection order.

megawebmaster:
Hi @hadihallak - do you have any news on the topic?

hadihallak (approved):
@megawebmaster We’ve been busy working on other issues and the future React 18 and concurrency work and haven’t had much time to look again into it until today.

I wasn’t able to run your reproduction repo as it was giving me a bunch of storybook/babel errors so I just tested it on a local app and I think the solution you worked on in this PR is good to make sure that ٍStitches isn’t causing any errors when it tries to hydrate the stylesheets by scanning them so this looks good to me and I’m gonna go ahead and merge it. Thank you so much for your patience and for working on a fix for it.

PR #856, opened 2021-10-22
perf: ⚡️ add dataset att to identify stitches tag

oeduardoal:
## This PR add a dataset attribute in style tag used by stitches to identify it.

`data-stitches=true`

### Result
image

> This will be useful to create a workaround to use tools like hotjar and FullStory that doesn't support the CSSOM changes.

> Example with FullStory:
> https://user-images.githubusercontent.com/12778450/138529841-a3b0c6ac-3683-4f7c-a211-d7ffade55098.mov

The workaround to implement when you use a tool to recording sessions:
https://gist.github.com/oeduardoal/499923b72422e4222c5073ba2a708ad1

### DIFF
--- a/packages/core/src/sheet.js
+++ b/packages/core/src/sheet.js
@@ -89,8 +89,14 @@ export const createSheet = (/** @type {DocumentOrShadowRoot} */ root) => {
 				})
 			}
 
+
+			const styleEl = document.createElement('style')
+			styleEl.dataset.stitches = 'true'
+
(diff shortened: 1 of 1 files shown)


hadihallak:
@oeduardoal Hey 👋 
thanks for looking into this and creating a PR but just so that i understand your case correctly.
you need the style tag to have unique identifier that you can use to search for it, correct?

because if that's the case, stitches internally has a unique structure to the sheet that you can use to search for the stitches populated sheet. like in this snippet:

```
const stitchesStyleTag = [...document.head.querySelectorAll('style')].find(tag => tag.sheet.cssRules[0].selectorText === '--sxs')
```

PR #893, opened 2021-12-12
878: Fixed issue with ThemeTokens typing

LucasUnplugged:
Simple fix to address the typing issue in #878; the offending type could be of types beyond `string | number`, so I added a type guard.

CC: @dlehmhus

Issue #749, opened 2021-08-30
TS error when using `as` + variant with composed components


peduarte:
To repro:

```tsx
const Box = styled('div', {
  variants: {
    size: {
      small: {}
    }
  }
});
const Flex = styled(Box, {})

export function Component() {
  return <Flex as="article" size="small" />;
}
```

jonathantneal:
Replication @ https://codesandbox.io/s/ecstatic-mayer-lobwy?file=/src/App.tsx

janhoogeveen:
If you want to update Stitches to the latest version but you're running into this issue... you can put // @ts-expect-error above the prop that's throwing the error and you can continue with your project.

KirdesMF:
maybe this is coming from the `as` prop types as `never` in the `styled-component.d.ts` .
I guess it could be type as `keyof JSX.IntrinsicElements` or something like that

csantos-nydig:
I have a scenario somewhat related:

```
const Box = styled('div', {
  color: 'red'
});

type MyBoxProps = {
  // --- how do I get the `as` prop type here? --- //
}

export function MyBox(props: MyBoxProps) {
  return <Flex as={props.as} size={props.size} />;
}
```

LucasUnplugged:
> type MyBoxProps = {
>   // --- how do I get the `as` prop type here? --- //
> }

Based on what I can see in the code, this should work:
```TSX
// Put this in your stitches.config.ts file
export type Polymorphic = string | React.ComponentType<any>

// In your components
type MyBoxProps = {
  as: Polymorphic,
}
```

Mind you, IIRC, React has a type for HTML element strings, but that may conflict with Stitches' types once you try to pass it in, I'm not sure. If anyone tries it out, please let us know how it went :)

ChrisDalley:
Hi @hadihallak - I'm still seeing Polymorphic `as` typescript errors when composing components even when using the latest `1.2.7-0` canary build (unless I'm misunderstanding how this should work with Stitches) - I've put together a code sandbox showing the issue here https://codesandbox.io/s/stitches-polymorphic-inheritance-typescript-error-d1xn1o?file=/src/Paragraph.component.tsx

hainessss:
Please reopen. This issue persists.

rntvicente:
Hi,

This issue persits with me.

**Configurations**
- next: 1.2.1.6
- react: 18.12.0
- react-dom: 18.12.0
- @stitches/react: 1.2.8

[...]

PR #900, opened 2021-12-15
749: Fixed polymorphic variants type issue

LucasUnplugged:
**NOTE**: PR #893 should be merged before this one; if that gets rejected, I can redo this one in a fresh branch`*`.

Tiny PR to address #749; I'm not sure if the typing is quite right, but based on what I'm seeing in the code, it should be. I've manually tested it where I'm seeing the issue, in my local development, and this resolved it, with no adverse side-effects (that I have seen).

---

`*` — I did this for my own convenience, to be able to create a build that has both of these fixes, which are disrupting my TS workflow. As long as #893 gets merged first, this should have no impact on the final result.

### DIFF
--- a/packages/core/types/stitches.d.ts
+++ b/packages/core/types/stitches.d.ts
@@ -204,7 +204,7 @@ type ThemeTokens<Values, Prefix> = {
 	[Scale in keyof Values]: {
 		[Token in keyof Values[Scale]]: ThemeUtil.Token<
 			Extract<Token, number | string>,
-			Values[Scale][Token] extends string | number ? Values[Scale][Token] : string,
+			Values[Scale][Token] & (string | number),
 			Extract<Scale, string | void>,
 			Extract<Prefix, string | void>
 		>
--- a/packages/core/types/stitches.d.ts
+++ b/packages/core/types/stitches.d.ts
@@ -204,7 +204,7 @@ type ThemeTokens<Values, Prefix> = {
 	[Scale in keyof Values]: {
 		[Token in keyof Values[Scale]]: ThemeUtil.Token<
 			Extract<Token, number | string>,
-			Values[Scale][Token],
+			Values[Scale][Token] extends string | number ? Values[Scale][Token] : string,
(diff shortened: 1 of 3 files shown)


packages/core/types/stitches.d.ts
```diff
@@ -204,7 +204,7 @@ type ThemeTokens<Values, Prefix> = {
 	[Scale in keyof Values]: {
 		[Token in keyof Values[Scale]]: ThemeUtil.Token<
 			Extract<Token, number | string>,
-			Values[Scale][Token],
+			Values[Scale][Token] extends string | number ? Values[Scale][Token] : string,
```
LucasUnplugged, line 207: NOTE: This change is just inherited from #893, and not directly related to this PR.

packages/react/types/stitches.d.ts
```diff
@@ -276,7 +276,7 @@ type ThemeTokens<Values, Prefix> = {
 	[Scale in keyof Values]: {
 		[Token in keyof Values[Scale]]: ThemeUtil.Token<
 			Extract<Token, number | string>,
-			Values[Scale][Token],
+			Values[Scale][Token] extends string | number ? Values[Scale][Token] : string,
```
LucasUnplugged, line 279: NOTE: This change is just inherited from #893, and not directly related to this PR.

LucasUnplugged:
Not sure why this isn't picking up that the base branch changed, @hadihallak 🙈 

Let me know if you'd rather a do a fresh rebase off canary, to only show this actual change.

csantos-nydig:
#893 is merged, could this one be merged now? Thanks

hadihallak:
Hey @LucasUnplugged Thanks for your effort working on this PR. I did some tests and found some cases where the behaviour of the polymorphic as changed so I dug deeper and was able to solve the issue from the area it originated from in https://github.com/modulz/stitches/pull/954

Thanks again for your effort and for helping me fix this issue 🙏 

Closing in favor of #954

Issue #650, opened 2021-06-21
Generate friendly class names


mike-shtil:
**Is your feature request related to a problem? Please describe.**
For better debugging, it's useful to quickly understand which component controls a given markup via friendly class names.

**Describe the solution you'd like**
Have an option to generate friendly class names in DOM:
`<div class="sxrb1g9">hello</div>` becomes `<div class="MyReactComponent__MyStyledComponent__sxrb1g9"hello</div>`.
Would work similar to the `displayName` option for **styled components**: https://styled-components.com/docs/tooling#better-debugging

Suggested syntax:
```
createCss({
  ...
  debugClassNames: true // or e.g. process.env.NODE_ENV !== 'production'
})
```

**Describe alternatives you've considered**
Alternative is to, in unfamiliar markup, hunt for the right component by tag names and trial and error 😒

**Additional context**
This feature exists in CSS modules, styled components and many other solutions, would dare calling it a standard.
I believe this part will need to be expanded a bit to accept new configs (and also get component names somehow)
https://github.com/modulz/stitches/blob/676eac2c37f50ec1598c4daba3cfcb35dca92287/packages/core/src/features/css.js#L81

peduarte:
This isn't something that we can offer out-of-the-box. In order to generate class names based on your variable names, it needs to happen at build time - or conveniently via babel. 

There is a babel package that does this for Stitches, but I haven't personally tried: https://www.npmjs.com/package/babel-plugin-transform-stitches-display-name

mike-shtil:
Yeah, I see the logic in what you're saying. Appreciate the link!

LucasUnplugged:
Commenting here because this is also something I'm looking for. While the babel package above is great when using React Dev Tools, in terms of inspecting the DOM itself, this feature request would be great!

@peduarte I'm having trouble getting the Stitches to build locally, but here is my suggested implementation for this:

<img width="1430" alt="Screen Shot 2021-07-22 at 1 39 35 PM" src="https://user-images.githubusercontent.com/2473823/126684088-2f293e12-4759-4f80-b364-cd9c4d250a7e.png">

If I'm understanding everything correctly, that means one could do the following:
```JSX
const Flex = styled('div', 'Flex', { display: 'flex' })
```
And that would result in a DOM output like this:
```HTML
<div class="my-prefix-c-Flex+ktIlUM">...</div>
```
(Obviously the `+` connector could be something else, like `_`)

What do you think of adding something like this?

[...]

PR #916, opened 2022-01-03
Fix for #650: Generate friendly class names

LucasUnplugged:
(Addresses #650)

I'm giving this one another try, with the latest version (plus my suggested fixes from PRs #893 and #900).

I've been working from a build off this branch for a couple of weeks now, with no adverse side-effects. This has resulted in this kind of rendered HTML:
```HTML
<section class="c-Box-lesPJm c-Flex-dhzjXW c-Column-iTKOFX c-Box-lesPJm-cScfiu-mb-6">
...
</section>
```

...which at this point, my brain quickly reads as (by ignoring the hashes):
```HTML
<section class="c-Box c-Flex c-Column c-Box-variant-mb-6">
...
</section>
```

I find that vastly more helpful, for debugging purposes, than a sea of hashed classes that mean nothing to me.

----

## Proposed API

Users can pass in a string as the last parameter of their calls to `styled`, like so:
```TSX
// Renders as `c-Label-sOm3h4Sh`
const Label = styled("label", {...}, "Label")
```

This proposed solution can be combined with [a babel plugin](https://www.npmjs.com/package/babel-plugin-named-stitches-classnames), to make friendly class names possible with no extra effort, for those using Babel.

----
## Related Discussion

There was a related discussion at #677, but the solution proposed there by the Stitches team (`defaultProps`) was removed from the v1 API.

### DIFF
--- a/packages/core/src/features/css.js
+++ b/packages/core/src/features/css.js
@@ -28,6 +28,10 @@ const createCssFunctionMap = createMemo()
 /** Returns a function that applies component styles. */
 export const createCssFunction = (/** @type {Config} */ config, /** @type {SheetGroup} */ sheet) =>
 	createCssFunctionMap(config, () => (...args) => {
+		const last = args.length - 1
+		const hasCustomName = last > 0 && args[last] && typeof args[last] === 'string'
+		const componentName = hasCustomName ? args[last].replace(/\./g, '_').replace(/\W/g, '') : null
+
 		/** @type {Internals} */
 		let internals = {
 			type: null,
@@ -53,8 +57,8 @@ export const createCssFunction = (/** @type {Config} */ config, /** @type {Sheet
 			}
 
 			// otherwise, add a new composer to this component
-			else {
-				internals.composers.add(createComposer(arg, config))
+			else if (typeof arg !== 'string') {
+				internals.composers.add(createComposer({componentName, ...arg}, config))
 			}
 		}
 
@@ -66,9 +70,11 @@ export const createCssFunction = (/** @type {Config} */ config, /** @type {Sheet
 	})
 
 /** Creates a composer from a configuration object. */
-const createComposer = (/** @type {InitComposer} */ { variants: initSingularVariants, compoundVariants: initCompoundVariants, defaultVariants: initDefaultVariants, ...style }, /** @type {Config} */ config) => {
+const createComposer = (/** @type {InitComposer} */ { variants: initSingularVariants, compoundVariants: initCompoundVariants, defaultVariants: initDefaultVariants, componentName, ...style }, /** @type {Config} */ config) => {
+	/** @type {string} */
+	const hash = componentName ? `${componentName}-${toHash(style)}` : toHash(style)
 	/** @type {string} Composer Unique Identifier. @see `{CONFIG_PREFIX}-?c-{STYLE_HASH}` */
-	const className = `${toTailDashed(config.prefix)}c-${toHash(style)}`
+	const className = `${toTailDashed(config.prefix)}c-${hash}`
 
 	/** @type {VariantTuple[]} */
 	const singularVariants = []
--- a/packages/core/src/features/css.js
+++ b/packages/core/src/features/css.js
@@ -72,9 +72,9 @@ export const createCssFunction = (/** @type {Config} */ config, /** @type {Sheet
 /** Creates a composer from a configuration object. */
 const createComposer = (/** @type {InitComposer} */ { variants: initSingularVariants, compoundVariants: initCompoundVariants, defaultVariants: initDefaultVariants, componentName, ...style }, /** @type {Config} */ config) => {
 	/** @type {string} */
-	const hash = componentName ? `${componentName}-${toHash(style)}` : toHash(style)
+	const baseClass = componentName && componentName.length > 0 ? `${componentName}-${toHash(style)}` : toHash(style)
 	/** @type {string} Composer Unique Identifier. @see `{CONFIG_PREFIX}-?c-{STYLE_HASH}` */
-	const className = `${toTailDashed(config.prefix)}c-${hash}`
+	const className = `${toTailDashed(config.prefix)}c-${baseClass}`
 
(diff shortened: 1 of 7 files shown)


emdowu:
why is this being blocked?

LucasUnplugged:
> why is this being blocked?

I'm not sure it's being blocked, so much as the Stitches team has been focused elsewhere; to my knowledge, it is not an intentional blocking.

luixo (commented):
Should we add tests for the feature?
I can take over a unit test.

LucasUnplugged:
> Should we add tests for the feature?
> 
> I can take over a unit test.

Oh, good point. I think I had done so already in my old commit; I'll see if I can find them and add them here. Thanks!

packages/core/src/features/css.js
```diff
@@ -66,9 +70,11 @@ export const createCssFunction = (/** @type {Config} */ config, /** @type {Sheet
 	})
 
 /** Creates a composer from a configuration object. */
-const createComposer = (/** @type {InitComposer} */ { variants: initSingularVariants, compoundVariants: initCompoundVariants, defaultVariants: initDefaultVariants, ...style }, /** @type {Config} */ config) => {
+const createComposer = (/** @type {InitComposer} */ { variants: initSingularVariants, compoundVariants: initCompoundVariants, defaultVariants: initDefaultVariants, componentName, ...style }, /** @type {Config} */ config) => {
+	/** @type {string} */
+	const hash = componentName ? `${componentName}-${toHash(style)}` : toHash(style)
```
luixo, line 75: I believe `componentName` is not a part of the hash.
It probably should be something like
```ts
const className = toTailHash(config.prefix) + ['c', componentName, toHash(style)].filter(Boolean).join('-');
```

LucasUnplugged: _If_ I understand you correctly (and I'm not sure I do), I think both pieces of code are doing the same thing; it's just my use of a variable name `hash` that is causing confusion.

My code is equivalent to doing this (note the change from `hash` to `baseClass`):
```JS
let baseClass = toHash(style)
if (componentName) {
  baseClass = `${componentName}-${baseClass}`
}
const className = `${toTailDashed(config.prefix)}c-${baseClass}`
```
So in both your code and mine, the `componentName` is getting conditionally added after the prefix and `c-`, but before the generated hash.

LucasUnplugged: But I'll change the name of the variable, to avoid confusion. Let me know if you're okay with that.

LucasUnplugged:
I've added some tests to both `core` and `react` packages, and changed that confusing variable name.

packages/core/src/features/css.js
luixo: Yep, they're doing the same thing, my note was about naming.
Thanks!

guipolitano:
Hope this PR gets merged soon

hadihallak:
> Hope this PR gets merged soon

@guipolitano Sadly, i don't think it's possible to get this merged as is as the API conflicts with another stitches API which is composing n number of styles into one component 
```
const Button = styled('button', {...styles}, {...morestyles}, anotherStyledComponent)
```

To be clear, We love this feature and we know how important it is for the reasons outlined in the PR but we just need to think more about the API in order to maintain the DX that everyone loves about Stitches


We will think more about this and hopefully come back with a plan/decision for the API sometime next week and discuss it with @LucasUnplugged and everyone in here.

LucasUnplugged:
> @guipolitano Sadly, i don't think it's possible to get this merged as is as the API conflicts with another stitches API which is composing n number of styles into one component 
> 
> ```
> 
> const Button = styled('button', {...styles}, {...morestyles}, anotherStyledComponent)
> 
> ```

@hadihallak  my code shouldn't conflict with that usage, because it only applies if there are at least 2 arguments, and the last argument is a string. In your example, only the first argument is a string, so it should still work as intended.

hadihallak:
That's true but it's re-using the same API which we've so far only reserved for styling so i'm wondering if we can do it differently so that it can be more clearly documented and easily readable for someone who isn't familiar with Stitches and reading the code for the first time

Maybe we could add a method or something similar that more explicitly tells the user what's going on behind the scenes.
We will think this through and get back and discuss things here with you

LucasUnplugged:
That makes sense, @hadihallak. Something like this would indeed be clearer:
```
const Button = styled('button', {...}).named('Button').
```

I'm not even sure if that's (realistically) possible, so didn't even attempt it.

guipolitano:
> > Hope this PR gets merged soon
> 
> @guipolitano Sadly, i don't think it's possible to get this merged as is as the API conflicts with another stitches API which is composing n number of styles into one component
> 
> ```
> const Button = styled('button', {...styles}, {...morestyles}, anotherStyledComponent)
> ```
> 
> To be clear, We love this feature and we know how important it is for the reasons outlined in the PR but we just need to think more about the API in order to maintain the DX that everyone loves about Stitches
> 
> We will think more about this and hopefully come back with a plan/decision for the API sometime next week and discuss it with @LucasUnplugged and everyone in here.

Sadly =/.
Maybe it wouldn't be better to use babel-plugin, just as Styled-Components do? I had try to create one, but i know nothing about it so I couldn't achieve it.

hadihallak:
@guipolitano 
yeah that's actually one of the main reason we need this feature. to enable us to automate the process of creating stable hashes and friendly classNames which is currently not possible as we have no API underneath to support it yet

guipolitano:
> @guipolitano yeah that's actually one of the main reason we need this feature. to enable us to automate the process of creating stable hashes and friendly classNames which is currently not possible as we have no API underneath to support it yet

Understood, hope you guys find a way. Thx for the great job, im loving to use Stitches with Radix. :)

hadihallak:
Closing in favor of #1005

bartcheers:
Closing the loop here: `withConfig` has been introduced after this issue was closed, achieving the desired behavior. It's still missing in the docs, but instructions can be found [here](https://github.com/stitchesjs/stitches/pull/1103).

PR #926, opened 2022-01-19
feat: Added static "as" method

cometkim:
This is reopened suggestion of #207 

The old PR was closed because incompatible with the new codebase, so I've reimplemented it to fit the current codebase.

I still think this is the simplest solution to get away with the ["as props nightmare"](https://twitter.com/jjenzz/status/1423766700885954562) and made PR directly to demonstrate just how small and simple the changes are.

### DIFF
--- a/packages/react/src/features/styled.js
+++ b/packages/react/src/features/styled.js
@@ -17,31 +17,37 @@ export const createStyledFunction = ({ /** @type {Config} */ config, /** @type {
 
 		const styled = (...args) => {
 			const cssComponent = css(...args)
-			const DefaultType = cssComponent[internal].type
 
-			const styledComponent = React.forwardRef((props, ref) => {
-				const Type = props && props.as || DefaultType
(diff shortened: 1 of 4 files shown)


arcc1845:
Progress of this PR?

LexSwed:
All the test cases can be covered through composition:

```ts
const Test = styled('div', bla)

const Link = styled('a', Test)

const Button = styled('button', Link)
```

So there is no need in this API

cometkim:
@LexSwed Oh thanks, I've never noticed that

I expect the API only change element, do not composing styles. for example:

```ts
const ButtonLink = Button.as(Link); // only use Button's style
```

`styled(Link, Button)` can be used for it?

hadihallak:
Will be closing this one as @LexSwed pointed this can already be done via composition so we'd prefer to use already existing APIs instead of adding new ones.

Regardless, Thank you so much for working on this @cometkim .

PR #929, opened 2022-01-20
improves compatibility with inconsistent CSSOM behavior of older browsers

cometkim:
Migrating from Emotion to Stitches, I ran into issues with an older iOS environment.

The cause is as described at https://github.com/modulz/stitches/issues/756

However, some old browsers assign null instead of throwing a `SecurityError` on the accessing `cssRules`.

Screenshot 2022-01-20 PM 7 04 39


hadihallak (approved):
LGTM. thanks for submitting this PR

Issue #908, opened 2021-12-27
Calling getCSSText returns [object Object] or [object Undefined] with Remix


joshuaellis:
Whilst I appreciate there's no formal documentation as of yet, I followed this [gist](https://gist.github.com/jjenzz/b67c8bd263bbf0f4e39c5fca55db93ee) as suggested by a fair few folk on the internet.

My attempt can be seen here: https://github.com/joshuaellis/remix-stitches

However, when you run that demo, you can see the SSR side of the styling works correctly, the `Hello World` is the specified red from the theme created by `createStitches`. When the client side hydrates, `getCssText` does not call the internal `toString` property you had written, it returns the native `object.toString` function, which therefore returns `[object Object]` or in some cases `[object Undefined]`.

What I believe to be happening during client hydration is `groupSheet` is becoming this object:

https://github.com/modulz/stitches/blob/c48a6532db350093dbd00fa2bace29930c53ba68/packages/core/src/sheet.js#L86

which evidently does not have a `toString` function. Patching this function does return the `theme` stylesheets, with all the colors applied to `:root` but then the text is still not "red" because the styles from the components and their class names have not been collected and put into the `cssRules` object.

Any help / guidance would be deeply appreciated!

JustinVoitel:
Same applies for SvelteKit applications!

I tried renaming the internal `toString` method of the sheet object to something else and it fixed the error in my case, so pretty trivial change imo.

jckw:
Interestingly, I'm finding that if I don't call `getCssText` in root.tsx or entry.server.tsx, this issue doesn't occur.

There are then other issues though:

1. the return value of `getCssText` differ on server and client because of new lines (which throws a console error - otherwise minor)
2. classnames generated by the `css()` and `styled()` funcs are not picked up on either the client or server output of `getCssText()`


---
Other potentially related things I've noticed:

3. calling `getCssText()` before the call to `renderToString` in entry.server.tsx gives a different output to calling it afterwards, namely missing css and styled generated classnames
4. calling `getCssText()` in the Document component also misses the classnames generated by css and styled (which can probably be explained via (3))

What confuses me is how the Next JS example works so simply, but Remix is more complex. I suppose I'd need to know more about the inner workings of both libraries to be able to work out why. 

I hope this helps!

joshuaellis:
[...]

PR #930, opened 2022-01-25
Fix for #908: toString in hydrated group sheet

luixo:
This PR addresses https://github.com/modulz/stitches/issues/908.

There was not tooling for tests with DOM API mocks, so I wrote a minimal one.


hadihallak (approved):
Thanks for debugging and fixing the issue

I've created separate issue to make sure that we have an official way of integrating with remix and another one to integrate testing with jsdom to make things easier to test things in a browser like env

PR #960, opened 2022-02-19
Fix CSS type for utils overshadowing properties

### LINKED ISSUE #921: Util types are shadowed by CSS properties, should be the other way around
# Bug report

## Describe the bug

When defining utils that have the same name as CSS properties, if I specify this util in the `css` or `styled` function then the util will be applied instead of the original CSS property, however it still expects the type of the original CSS property.
For example, say I define a util called `padding` that only accepts arrays of numbers as it's argument. Now when I add the padding property to the style object I provide to the `styled` function, it will apply the util instead of the actual `padding` CSS property, but it will show the type of the CSS property, i.e. `OnlyStringNumeric | Globals | ScaleValue | Index | undefined` instead of the type I defined for the util which is `number[]`.

## To Reproduce

https://codesandbox.io/s/stitches-bug-utils-type-shadowing-h3x2u?file=/src/App.tsx

## Expected behavior

I expect the `css` and `styled` functions to show the type of the util and not the type of the shadowed CSS property, since the util is what's being applied.

## System information

- OS: Windows
- Browser (if applies) chrome,
- Version of Stitches: 1.2.6

mszczepanczyk:
When util's and property's names clash the util's type should take precedence.

Fixes #921

BTW type testing doesn't seem to work, the tests will pass regardless of the typing.

### DIFF
--- a/packages/react/types/css-util.d.ts
+++ b/packages/react/types/css-util.d.ts
@@ -27,7 +27,7 @@ export type CSS<
 	}
 	// known property styles
 	& {
-		[K in keyof CSSProperties]?: (
+		[K in keyof CSSProperties as K extends keyof Utils ? never : K]?: (
 			| ValueByPropertyName<K>
 			| TokenByPropertyName<K, Theme, ThemeMap>
 			| Native.Globals
--- a/packages/core/tests/issue-921.ts
+++ b/packages/core/tests/issue-921.ts
@@ -1,0 +1,62 @@
+import { createStitches, PropertyValue, CSS } from '../types/index'
+
+const config = {
+	utils: {
+		background: (value: boolean | PropertyValue<'background'>) => {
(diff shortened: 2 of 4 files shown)


packages/react/types/css-util.d.ts
```diff
@@ -27,7 +27,7 @@ export type CSS<
 	}
 	// known property styles
 	& {
-		[K in keyof CSSProperties]?: (
+		[K in keyof CSSProperties as K extends keyof Utils ? never : K]?: (
```
hadihallak, line 30: I remember that at some point, we explicitly didn’t want utils to override CSS properties, and the types for utils below currently showcase the intention, so this PR doesn’t completely fix the issue. 

Now changing this runtime behavior to match the types would be a breaking change to existing users, so I will go ahead and merge this and then fix the type for utils so that it matches what happens at runtime

PR #992, opened 2022-04-04
Format code with Prettier

lesha1201:
I noticed that some code wasn't formatted. I thought to run an existing script to format all the code but haven't found it so I added it to the root `package.json`:

image

I also changed a print width to 80 to increase readability. Previous value `240` were producing not very readable code in some places.

ESLint doesn't pass now because it's not configured. We should add configuration for ESLint which uses `eslint-config-prettier`.

I would check the format in CI but I haven't found a workflow for linting so I omitted it for now. I think we should create a workflow that lints and checks the format.


hadihallak:
Hey @lesha1201 
Thanks for working on this. the codebase indeed needs some prettier love but since this PR creates such a huge diff, I've created an issue to track it and include it as part of the next milestone.

For now, I'm gonna close this issue.

Again, Thank you so much for contributing to Stitches

PR #998, opened 2022-04-10
Fix missing types

BESHOI:
add types to size-adjust, ascent-override;

### DIFF
--- a/packages/react/types/css.d.ts
+++ b/packages/react/types/css.d.ts
@@ -9303,6 +9303,8 @@ export namespace AtRule {
     fontWeight?: FontWeight;
     src?: string;
     unicodeRange?: string;
+		sizeAdjust?: string;
+  	ascentOverride?: string;
   }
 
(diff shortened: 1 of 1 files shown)


packages/react/types/css.d.ts
```diff
@@ -9303,6 +9303,8 @@ export namespace AtRule {
     fontWeight?: FontWeight;
     src?: string;
     unicodeRange?: string;
+		sizeAdjust?: string;
+  	ascentOverride?: string;
```
LucasUnplugged, line 9307: This was probably accidental, but you should try to maintain the existing indentation.

BESHOI: oh, thanks i will

hadihallak (approved):
Thanks for the PR @BESHOI ❤️

BESHOI:
Thank you so much @hadihallak 😍

Issue #832, opened 2021-10-06
Stitches throwing exceptions "TypeError - Converting circular structure to JSON"


YuvalZiegler:
# Bug report

We use stitches at [attentive](https://www.attentivemobile.com) to power our internal component library and style system, we love it, so first off, thank you! Part of what we do at attentive is render UI on our clients' sites, we recently switched to using our new internal library, powered by stitches, and ran into this circular dependency issue.

## Describe the bug

When any circular dependency exists in the document, `createStitches` fails

## To Reproduce

I've created a code sandbox that reproduces the issue here: https://codesandbox.io/s/circular-json-reproduction-0i5lp?file=/src/App.tsx

to reproduce locally:

```js
document.someLib = {};
document.someLib.ref = someLib;

const { styled } = createStitches({});
// ^ throws  "TypeError - Converting circular structure to JSON"
```
## Expected behavior

Should be able to handle circular JSON

## Screenshots

with a circular dep: 

<img width="1421" alt="Screen Shot 2021-10-06 at 10 36 24 AM" src="https://user-images.githubusercontent.com/248544/136224971-acfc6690-38f0-4b75-a898-923170cb46c9.png">

without

<img width="915" alt="Screen Shot 2021-10-06 at 10 37 35 AM" src="https://user-images.githubusercontent.com/248544/136225188-6d8bb938-6b2e-4217-b104-397b1168ed66.png">

## System information

- OS: macOS (but will affect all)
- Browser: All
- Version of Stitches: 1.2.3

## Note

Related to this issue was recently closed: https://github.com/modulz/stitches/issues/665#issue-941848074

peduarte:
Hey, I created a PR (https://github.com/modulz/stitches/pull/836) 

I've forked your sandbox, and updated the dependency to use the build from that PR.

Here's a sandbox: https://codesandbox.io/s/circular-json-reproduction-forked-w3cr5?file=/package.json

Looks like that fixes the problem

YuvalZiegler:
this is great! thanks for the fast fix!

PR #1004, opened 2022-04-13
Add `root` to `config` (#832)

jlalmes:
Closes https://github.com/modulz/stitches/issues/832.

> **`root` was incorrectly removed from config here: https://github.com/modulz/stitches/pull/836**

## Add `root` to `config`

The `root` config property is causing a `TypeError - Converting circular structure to JSON` error. The root cause of this is the `createMemo` function (introduced [here](https://github.com/modulz/stitches/issues/625)) attempting to `JSON.stringify` circular objects such as `document`.

### Changes made by this PR

* Added better `stringify` function used by `createMemo.js` (heavily inspired by [`safe-json-stringify`](https://www.npmjs.com/package/safe-json-stringify))
* Added `root` to `CreateStitches` & `ConfigType` TS definitions (for `core` & `react`).
* Added new regression test for [#832](https://github.com/modulz/stitches/issues/832)

### DIFF
--- a/packages/core/src/createStitches.js
+++ b/packages/core/src/createStitches.js
@@ -34,6 +34,7 @@ export const createStitches = (config) => {
 			theme,
 			themeMap,
 			utils,
+			root,
 		}
 
 		/** Internal stylesheet. */
--- a/packages/core/src/utility/createMemo.js
+++ b/packages/core/src/utility/createMemo.js
@@ -1,13 +1,13 @@
-const stringifyReplacer = (name, data) => (typeof data === 'function' ? { '()': Function.prototype.toString.call(data) } : data)
+import { safeJsonStringify } from './safeJsonStringify.js'
 
-const stringify = (value) => JSON.stringify(value, stringifyReplacer)
+const stringifyReplacer = (name, data) => (typeof data === 'function' ? { '()': Function.prototype.toString.call(data) } : data)
 
 /** @type {() => <T = any, A = any[], F = (T, ...A) => any>(value: T, apply: F, ...args: A) => ReturnType<F>} */
 export const createMemo = () => {
 	const cache = Object.create(null)
 
 	return (value, apply, ...args) => {
-		const vjson = stringify(value)
+		const vjson = safeJsonStringify(value, stringifyReplacer)
 
 		return vjson in cache ? cache[vjson] : (cache[vjson] = apply(value, ...args))
 	}
--- a/packages/core/src/utility/safeJsonStringify.js
+++ b/packages/core/src/utility/safeJsonStringify.js
@@ -1,0 +1,63 @@
+// https://github.com/debitoor/safe-json-stringify/blob/master/index.js
+import { hasOwn } from './hasOwn.js'
+
+function throwsMessage(err) {
+	return '[Throws: ' + (err ? err.message : '?') + ']'
+}
+
+function safeGetValueFromPropertyOnObject(obj, property) {
+	if (hasOwn(obj, property)) {
+		try {
+			return obj[property]
+		} catch (error) {
+			return throwsMessage(error)
+		}
+	}
+	return obj[property]
+}
+
+function ensureProperties(obj) {
+	const seen = [] // store references to objects we have seen before
+
+	function visit(obj) {
+		if (obj === null || typeof obj !== 'object') {
+			return obj
+		}
+
+		if (seen.indexOf(obj) !== -1) {
+			return '[Circular]'
+		}
+
+		seen.push(obj)
+
+		if (typeof obj.toJSON === 'function') {
+			try {
+				const result = visit(obj.toJSON())
+				seen.pop()
+				return result
+			} catch(err) {
+				return throwsMessage(err)
+			}
+		}
+
+		if (Array.isArray(obj)) {
+			const result = obj.map(visit)
+			seen.pop()
+			return result
+		}
+
+		const result = Object.keys(obj).reduce(function(acc, prop) {
+			// prevent faulty defined getter properties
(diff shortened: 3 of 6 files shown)


jlalmes:
An alternate solution might be removing the stringify in `createMemo`.

```javascript
export const createMemo = () => {
	const cache = new Map()

	return (value, apply, ...args) => {
                 if (cache.has(value)) {
                          return cache.get(value)
                 }
                 const result = apply(value, ...args)
                 cache.set(value, result)
		return result
	}
}
```

@jonathantneal could you give some feedback on this?

aMediocreDad:
If it is of interest I wrote a different solution last year (it was for use in a personal project as it uses the constructed stylesheets polyfill). https://github.com/aMediocreDad/stitches/commit/14660d439d66086f9d304a714dd1bb9d4c33b782

cinchy-m:
@jlalmes Any chance you could add core lib usage instructions for web components to the docs in this PR?

jlalmes:
Happy to update the Stitches website docs once this PR is merged 👍

LexSwed:
Hey @hadihallak 👋 
Any chance `stitches` usage with Shadow DOM can be included in the next [Milestone](https://github.com/stitchesjs/stitches/milestone/1)? The issue I'm looking to solve is the ability to pass the `root` to the `createStitches` config, so that I do not polute user page styles with `stitches` properties when developing browser extension with a content script that renders UI in own shadow root.

hadihallak:
@jlalmes Thanks for your hard work.. this is something that's under consideration for the next release.
I will be reviewing this in the next couple of days but giving this a quick look it looks good but I'm worried about the added 
bundle size so I'm wondering if we can do it some other / more efficient way.

@LexSwed Thanks for the ping
Did you get the chance to test this PR in your setup? have you had any issues?

LexSwed:
> Did you get the chance to test this PR in your setup? have you had any issues?

No, I haven't tried this solution specifically, but this one: https://github.com/stitchesjs/stitches/issues/628#issuecomment-1074327282 worked in my case

jlalmes:
Hi @hadihallak - I suggest using a better memoizing function like the one I have mentioned above if bundle size is a concern (https://github.com/stitchesjs/stitches/pull/1004#issuecomment-1097783207). Let me know what you think and I will update the PR.

hadihallak:
@jlalmes @LexSwed I think we're gonna delay merging this until after the 1.3 release as even with root support via the config, this doesn't solve providing root at runtime like in other libraries so I'm gonna think about this after the 1.3 release and come up with a plan to include runtime root support or work on merging this with the same old root API that exists here.

Just to be extra clear, we want to support this as it's a much needed feature even for things not mentioned in this PR but we don't want to be hasty with API decisions (I know this used to exist prior to v1 but it wasn't the ideal solution).

Thank you so much for your hard work 🙏

IsLand-x:
Need shadow DOM feature support.

When creating chrome extension and injecting ui into content script, this "css in js" library will be the best practice if shadow dom feature was supported.

precious-void:
Any updates on the status of PR?

he-la:
Don't we also have to add the `root` type to the exported config type in `stitches.d.ts`?

```patch
From: Henrik Laxhuber <henrik@laxhuber.com>
Date: Wed, 22 Mar 2023 20:09:29 +0100
Subject: [PATCH] Shadow DOM Support

---
 packages/core/types/stitches.d.ts  | 4 +++-
 packages/react/types/stitches.d.ts | 4 +++-
 2 files changed, 6 insertions(+), 2 deletions(-)

diff --git a/packages/core/types/stitches.d.ts b/packages/core/types/stitches.d.ts
index 42a7200..3ed6636 100644
--- a/packages/core/types/stitches.d.ts
+++ b/packages/core/types/stitches.d.ts
@@ -12,7 +12,8 @@ export default interface Stitches<
 	Media extends {} = {},
 	Theme extends {} = {},
 	ThemeMap extends {} = {},
-	Utils extends {} = {}
+	Utils extends {} = {},
+	Root extends DocumentOrShadowRoot = Document
 > {
 	config: {
 		prefix: Prefix
@@ -20,6 +21,7 @@ export default interface Stitches<
 		theme: Theme
 		themeMap: ThemeMap
 		utils: Utils
+		root?: Root
 	},
 	prefix: Prefix
 	/** The **prefix** property defined.
diff --git a/packages/react/types/stitches.d.ts b/packages/react/types/stitches.d.ts
index 846f281..9d7b986 100644
--- a/packages/react/types/stitches.d.ts
+++ b/packages/react/types/stitches.d.ts
@@ -12,7 +12,8 @@ export default interface Stitches<
 	Media extends {} = {},
 	Theme extends {} = {},
 	ThemeMap extends {} = {},
-	Utils extends {} = {}
+	Utils extends {} = {},
+	Root extends DocumentOrShadowRoot = Document
 > {
 	config: {
 		prefix: Prefix
@@ -20,6 +21,7 @@ export default interface Stitches<
 		theme: Theme
 		themeMap: ThemeMap
 		utils: Utils
+		root?: Root
 	},
 	prefix: Prefix
 	/** The **prefix** property defined.
--
2.38.3
```

he-la:
While at it, might be useful to also support multiple roots + changing roots dynamically? If you have a need for Shadow DOM, then chances are you'll have more than one Shadow DOM.

maylynn-ng:
Hii!! Running into this issue, this would be an awesome addition! Any updates? 🙏

maylynn-ng:
> While at it, might be useful to also support multiple roots + changing roots dynamically? If you have a need for Shadow DOM, then chances are you'll have more than one Shadow DOM.

If this PR is looking good for a single one, we could maybe do that as a follow up PR to unblock some use cases? :)

Issue #1021, opened 2022-05-11
gridGap property incorrect type


panzelva:
# Bug report

## Describe the bug

When trying to use `gridGap: theme.space.someSpace`, TypeScript throws error:

```
Type 'Token<"someSpace", string, "space", "">' is not assignable to type 'Globals | Index | undefined'.
  Type 'Token<"someSpace", string, "space", "">' is not assignable to type 'number & Record<never, never>'.
    Type 'Token<"someSpace", string, "space", "">' is not assignable to type 'number'.ts(2322)
```

To workaround, user can use `columnGap` + `rowGap` (both work fine).

## To Reproduce

See this sandbox: https://codesandbox.io/s/naughty-pond-1rvogt?file=/src/App.tsx

## Expected behavior

`gridGap` does not throw TS error.

## System information

- OS: macOS
- Version of Stitches: 1.2.8
- Version of Node.js: 16

hadihallak:
@guilhermedeandrade is right. the two types are simply incompatible.
for the recommended way to do theming, please reference the [docs](https://stitches.dev/docs/tokens)

panzelva:
> I think this isn't a bug actually. What you get from `theme.space.someSpace` isn't a `string`, but actually a `Token`. You can access the value of it using the `value` property, so that would be `theme.space.someSpace.value`.
> 
> Here is the forked version from your Codesandbox: https://codesandbox.io/s/modulz-stitches-gridgap-property-incorrect-type-1021-forked-05sb6x?file=/src/App.tsx

That seems wierd to me, because in [docs](https://stitches.dev/docs/typescript#using-tokens) there is section with tips for stricter Typescript experience, and using `Token` directly is there as option.

And it works for every other property other than `gridGap`, so it looks like to me this is indeed a bug (or not ideal documentation).

guilhermedeandrade:
> > I think this isn't a bug actually. What you get from `theme.space.someSpace` isn't a `string`, but actually a `Token`. You can access the value of it using the `value` property, so that would be `theme.space.someSpace.value`.
> > Here is the forked version from your Codesandbox: [codesandbox.io/s/modulz-stitches-gridgap-property-incorrect-type-1021-forked-05sb6x?file=/src/App.tsx](https://codesandbox.io/s/modulz-stitches-gridgap-property-incorrect-type-1021-forked-05sb6x?file=/src/App.tsx)
> 
[...]

PR #1042, opened 2022-06-14
fix gridGap property incorrect type #1021

BESHOI:
add typescript support to grid-gap property


hadihallak (commented):
Could you please add the changes to the core package too?

BESHOI:
Sure

hadihallak (approved):

PR #1051, opened 2022-07-01
Use `unitlessProps` table instead of `unitProps` table to save bundle size

onurkerimov:
- While converting numeric values to pixel values, some bundle size can be saved if a blacklist is used, instead of a whitelist. Because when only numeric values are considered, unitless attributes are a lot less then props with units. 
- This is how emotion tackles the same conversion. It uses a package called `@emotion/unitless`, it was used as the main source in this PR. ([source](https://github.com/emotion-js/emotion/blob/main/packages/unitless/src/index.js))


packages/core/src/convert/toCssRules.js
```diff
@@ -153,136 +153,52 @@ const toCssString = (/** @type {string[]} */ declarations, /** @type {string[]}
 )
 
 /** CSS Properties whose number value may safely be interpretted as a pixel. */
```
onurkerimov, line 155:
```suggestion
/** CSS Properties whose number values should be unitless. */
```

hadihallak (approved):

PR #1082, opened 2022-09-06
fix(#1078): TS 4.8 compat

### LINKED ISSUE #1078: TypeScript issue with PropertyValue and TS v4.8.2
# Bug report

## Describe the bug

using `PropertyValue` for defining a custom util in TypeScript 4.8.2 generates the following typescript error:

```
TS4118: The type of this node cannot be serialized because its property '[$$PropertyValue]' cannot be serialized.
```

## To Reproduce

I cannot reproduce this in codesandbox because I can't change the TypeScript version (See: https://github.com/codesandbox/codesandbox-client/issues/942)

1. Create a react server with typescript 4.8.2
2. Create stitches config

```
const stitches = createStitches({
  utils: {
    marginHorizontal: (value: PropertyValue<'margin'>) => ({
      marginLeft: value,
      marginRight: value
    })
  }
});

export const { css, config, createTheme, getCssText, globalCss, keyframes, reset, styled, theme } =
  stitches;
```

3. notice the TS error

## Expected behavior

I should be able to use `PropertyValue` on TS 4.8.2

## Screenshots

Screen Shot 2022-09-02 at 10 19 26 AM

## System information

- OS: N/A
- Browser: N/A
- Version of Node.js: N/A
- Version of Stitches: 1.2.8
- Version of TypeScript: 4.8.2

jgoz:
Fixes #1078.

Something changed between TS 4.7 and 4.8 that prevents unique symbols from being considered serializable in the type system. I don't know if this change was deliberate or if the previous behavior was incorrect, but I couldn't find any TypeScript issues/PRs that were directly related.

I applied a workaround from https://github.com/microsoft/TypeScript/issues/37888#issuecomment-846638356, which avoids direct serialization of `$$PropertyValue` and `$$ScaleValue` by writing them as imports instead. Based on the type tests, it seems like this didn't break anything, but it's possible there are second order effects beyond what is being tested inside the this project.

It might be worth logging a TypeScript issue if we feel the previous behavior was correct and desirable, if only to get an answer from the team. In the meantime, this PR might unblock upgrades.


ivanbanov:
@hadihallak any chance that we get it out of the door? 🙏 🥺

vorillaz:
Hi all, 
I honestly don't want to be this kind of person but this PR seems like a deal breaker for people using stitches or willing to start using the project.

Hope it will get merged soon, thanks a ton for your wonderful work.

jgoz:
Btw, I logged a TypeScript bug for this issue, but no guarantees that it will be fixed, so this workaround might continue to be relevant: https://github.com/microsoft/TypeScript/issues/50720

package.json
```diff
@@ -32,11 +32,11 @@
     "@radix-ui/react-separator": "0.1.0",
     "@skypack/package-check": "0.2.2",
     "@types/node": "16.9.6",
-    "@types/react": "17.0.33",
```
hadihallak, line 35: Is this change required? i'd prefer not to update anything to react 18 as this might signal that the library is fully compatible with it

jgoz: I don’t recall. I’ll check today.

hadihallak: @jgoz Thank you so much!

jgoz: Reverted. Thanks for calling this out. I forget why I did this originally &mdash; maybe I thought the error was related to react's types in some way. But it wasn't 😄

hadihallak (approved):

linuz90:
Any update on this? Using Stitches for Typefully and can't find a way to remove the errors and build the project now.

I'd even install the Stitches canary if possible.

hadihallak:
@linuz90 
I think we're going to release a beta around early next week. 
If you're in a rush then I'd suggest you build it and then publish to your personal npm account as a temporarily solution

linuz90:
I think I'll try with the Typescript downgrade until then, glad to hear a beta is coming soon 🙌

gajus:
I am assuming this was not released?

jgoz:
It's in the latest beta releases.

PR #1083, opened 2022-09-08
feature(token type): add opacities as new supported token type

ioanatatu:
## Small feature proposal

The project I am working on requires opacities as a token type, which is currently not among the 14 token types supported by Stitches. While going through the issues I also noticed there is one other [mention](https://github.com/stitchesjs/stitches/issues/1043#issue-1271437493) of opacities, but the issue is not actually a feature proposal.


hadihallak:
Hey @ioanatatu 

Thank you so much for working on this issue and correctly updating the types package 😄 

As I mentioned in the [issue](https://github.com/stitchesjs/stitches/issues/1043#issuecomment-1270385470) that you linked, this already works in Stitches but autocompletion does not work for tokens that are part of the newly added opacities scale

I've tagged the original issue under `typescript` and once that is fixed then autocompletion should work for new token scales without modifying the core Stitches code.

Again, Thank you so much for working on this but for now i'm going to close this PR as this needs to be fixed at the types level.

PR #1086, opened 2022-09-15
add .js extension to .d.ts files (TS NodeNext fix)

### LINKED ISSUE #1084: `styled`, etc from `createStitches` are `any` (pnpm, esm, monorepo)
# Bug report

## Describe the bug

Using ESM / "Node next" module resolution, I'm encountering issues where TS knows how to type `createStitches`, but fails to get the typings of any of the produced tools.

I _believe_ this is due to using `moduleResolution: 'node16'` in TypeScript and/or `"type": "module"` in my app's `package.json`. Might be related to how that works with other tsconfig options.

I have a minimal repro which produces the problem locally here: https://github.com/a-type/stitches-esm-repro.

If you click through to the definition of `createStitches`, then follow up to the referenced `StitchesConfig` defintion, VS Code's Typescript can't follow the module reference.

TS can't follow the import of `./config` from `@stitches/react/types/index.d.ts`. Since it can't reach that file, it can't provide typings for any of my components using `styled`, etc.

I believe this could be because with ESM rules it expects `./config.js` as the module path, since file extensions are now mandatory.

My tsconfig is roughly:

```
{
  "compilerOptions": {
    "target": "es2017",
    "useDefineForClassFields": true,
    "lib": ["DOM", "DOM.Iterable", "ESNext"],
    "allowJs": false,
    "skipLibCheck": true,
    "esModuleInterop": true,
    "allowSyntheticDefaultImports": true,
    "strict": true,
    "forceConsistentCasingInFileNames": true,
    "module": "ES2022",
    "moduleResolution": "Node16",
    "resolveJsonModule": true,
    "isolatedModules": true,
    "noEmit": true,
    "jsx": "react-jsx",
    "rootDir": ".",
    "baseUrl": "."
  },
  "include": ["src"],
  "references": [{ "path": "./tsconfig.node.json" }]
}
```

In my main app it's more complex, but this seems to reproduce the problem.

## To Reproduce

1. Clone https://github.com/a-type/stitches-esm-repro
2. `pnpm install`
3. Open `src/App.tsx`
4. Check TS typing of `Button` or `styled`. My editor says `any`.

## Expected behavior

`styled`, etc are typed correctly and produce fully typed components.

## Screenshots

err

## System information

- OS: Windows 11
- Browser (if applies) N/A
- Version of Stitches: 1.2.8
- Version of Node.js: 16.13.0
- Typescript: 4.8.2

a-type:
fixes type module resolution for NodeNext/ESM in TypeScript

fixes #1084 

TypeScript requires `.d.ts` files to follow ESM module import rules as well, so these files needed extensions added to imported local modules so that types can be resolved when Stitches is used in a project that has `moduleResolution: 'NodeNext'`.

### DIFF
--- a/packages/core/types/config.d.ts
+++ b/packages/core/types/config.d.ts
@@ -1,5 +1,5 @@
-import type * as CSSUtil from './css-util'
-import type Stitches from './stitches'
+import type * as CSSUtil from './css-util.js'
+import type Stitches from './stitches.js'
 
 /** Configuration Interface */
 declare namespace ConfigType {
--- a/packages/core/types/css-util.d.ts
+++ b/packages/core/types/css-util.d.ts
@@ -1,7 +1,7 @@
-import type * as Native from './css'
-import type * as Config from './config'
-import type * as ThemeUtil from './theme'
-import type * as Util from './util'
+import type * as Native from './css.js'
+import type * as Config from './config.js'
+import type * as ThemeUtil from './theme.js'
+import type * as Util from './util.js'
 
 export { Native }
 
--- a/packages/core/types/index.d.ts
+++ b/packages/core/types/index.d.ts
@@ -1,10 +1,10 @@
-import type Stitches from './stitches'
+import type Stitches from './stitches.js'
 
-import type * as Config from './config'
-import type * as CSSUtil from './css-util'
-import type * as StyledComponent from './styled-component'
+import type * as Config from './config.js'
+import type * as CSSUtil from './css-util.js'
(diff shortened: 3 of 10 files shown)


a-type:
Well, I thought this would be enough, but using linking to test this locally all my styled native elements are being typed as `StyledComponent<string, ...` and are still broken.

a-type:
Yeah, there are minor ESM differences like not being able to import directory indexes which could trip up in the future, but hopefully this is good enough for now. Glad I can drop my local npm links now! Thanks for the review.

PR #1101, opened 2022-10-04
Revert "Fix gridGap property"

hadihallak:
Reverts stitchesjs/stitches#1099


hadihallak:
https://github.com/stitchesjs/stitches/pull/1099. was referencing main instead of canary

PR #1110, opened 2022-10-26
fix defaultThemeMap: add accentColor to default to support colors

xDellTog:
Hi there! I don't know if that's an issue or a possible feature, but I will consider an issue. 

Recently I used `accentColor` in my theme, to customize a native checkbox and **Stitches** don't apply the color in final CSS, so I added `accentColor` in `defaultThemeMap` to property support theme colors and it works. 
I used [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/CSS/accent-color) syntax and description to doc.

✅ **Works**
```
accentColor: 'red'
```

❌ **Failure**
```
accentColor: '$primaryColor'
```

[Sandbox to Test](https://codesandbox.io/s/accent-color-stitches-r7091r)

I hope it can be fixed, Thanks!


prclm:
could you remove all the auto formatting changes?

PR #1115, opened 2022-11-09
fix(package.json): Add 'types/*' to package exports

jeffberry:
This fixes an issue that we are experiencing when setting `moduleResolution` in our tsconfig to `NodeNext`. When this is enabled, the module resolution will only allow importing from paths that are defined within the `exports` config. Because only `types/index.d.ts` is available for import, it is causing these kinds of embedded imports to fail:

```
import("@stitches/react/types/css-util").CSS
```

I think the alternative would be to ensure that all types are exported from the `types/index.d.ts` file.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -21,6 +21,9 @@
       "import": "./dist/index.mjs",
       "types": "./types/index.d.ts"
     },
+    "./types/*": {
+      "types": "./types/*.d.ts"
+    },
     "./global": "./dist/index.global.js"
   },
   "files": [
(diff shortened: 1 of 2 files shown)


Achaak:
I confirm this need on my side to pass a lib in ESM and pass the moduleResolution in Node16

yovanoc:
Please accept this PR

smol-honk:
@peduarte who should we @ to get this looked at? Thank you again for your time! Love the library.

jonathantneal:
This actually looks really good. If I can get some access, I could merge this and put it into a release.

peduarte:
cc @hadihallak @StephenHaney

fmal:
@hadihallak @StephenHaney could you please merge this in and put it into a canary release or give @jonathantneal access do do so? This doesn't seem like much effort and would unblock usage of stitches with TypeScript's `moduleResolution` other than `node`. I would really appreciate it ☮️

PR #1150, opened 2023-04-03
Fixed `package.json#exports.types`

Andarist:
When relying on `exports.types` for loading types, this condition has to come first. `exports` are order-dependent - tests/loaded from the top to the bottom

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -2,7 +2,6 @@
   "name": "@stitches/core",
   "version": "1.3.1-1",
   "description": "The modern CSS-in-JS library",
-  "type": "module",
   "main": "dist/index.cjs",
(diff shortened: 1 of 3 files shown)


packages/core/package.json
```diff
@@ -2,7 +2,6 @@
   "name": "@stitches/core",
   "version": "1.3.1-1",
   "description": "The modern CSS-in-JS library",
-  "type": "module",
```
Andarist, line 5: I don't think that you actually needed this. You still provide `exports.import` after all. Having those packages as modules changes how `.js` (and this `.d.ts`) files are treated and supporting types with this setup would be much more complicated. So I think that there is no harm if we just remove this from here to make `types` work.

